PR #4, opened 2023-03-17
fix: support japanese typing

PeterTakahashi:
# bug
The Japanese language has both kanji and hiragana, and the enter key is used to convert hiragana into kanji.
In the current implementation, `handleSend` is executed when the conversion is confirmed and starts sending messages.

# solution
When the conversion key is pressed, the message is not sent.

スクリーンショット 2023-03-18 8 30 03


mckaywrigley:
@PeterTakahashi looks wonderful. Thank you for helping support Japanese users!

PR #7, opened 2023-03-18
Show the sidebar toggle button only on mobile

nauxliu:


mckaywrigley:
The way the UI will eventually unfold will make more sense with sidebar closing, so going to keep it for now.

PR #9, opened 2023-03-18
Ignore new message when the current message is not finished yet

nauxliu:


mckaywrigley:
Works great. Thank you for your contributions - they are excellent!

PR #11, opened 2023-03-18
Updated ChatInput.tsx text area to autoextend for larger prompt size

asolpshinning:
Made it similar to openAI ChatGPT text area behavior.


mckaywrigley:
Looks great - thanks!

PR #14, opened 2023-03-19
fixed chat textarea overflow

HaithamLeo:
Hi folks,

I'm a UI/UX developer.
This is my first contribution to this repo..tiny change but an important one :)

Hiding the overflow-y in a textarea can be beneficial for the UI/UX because it provides a cleaner and more consistent user interface. When the overflow-y is hidden, the textarea will only display the visible text and will not show any scrollbars or other elements that could clutter the interface.

This can be particularly useful when designing interfaces for mobile devices or small screens, where space is at a premium. By hiding the overflow-y, you can ensure that the textarea takes up only the necessary space and does not interfere with other elements on the page.

In addition, hiding the overflow-y can also help to prevent distractions and improve the user's focus on the content they are typing. When scrollbars or other elements are visible, they can be a distraction and may draw the user's attention away from the task at hand.

### DIFF
--- a/components/Chat/ChatInput.tsx
+++ b/components/Chat/ChatInput.tsx
@@ -13,6 +13,8 @@ export const ChatInput: FC<Props> = ({ onSend, messageIsStreaming }) => {
 
   const textareaRef = useRef<HTMLTextAreaElement>(null);
 
+  const chatBoxMaxHeight = 400; //chat box max height in pixels
+
   const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
     const value = e.target.value;
     if (value.length > 4000) {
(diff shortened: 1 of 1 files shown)


nauxliu:
If the message is too long, there is no way to scroll up and modify the start of the message. This is not a good UX experience. The approach of the official ChatGPT website is to provide a scrollbar when the message is too long.

HaithamLeo:
Good point, my bad ..I skipped that part

Then may be we could hide the the scroll bat til some height, ex: `200px` then show the scroll bar.

Also we can change the scroll bar to be more user appealing, like Chatgpt.

mckaywrigley:
Currently the input height maxes out at 400px. We could add the scroll bar once it hits that height.

HaithamLeo:
@mckaywrigley just updated my PR.
Now the scroll bar will be hidden until content exceeds more than `400px` height then the scroll bar will be visible.

mckaywrigley:
I'm going to close this for now just because it introduces a bit more issues than it resolves. Tested it and we lose a few pieces of UX like the textarea shrinking appropriately when the length shortens. Textarea height stuff is a bit finicky, but we can revisit this later.

PR #15, opened 2023-03-19
Correct the improper context being used when it exceeds the limit.

nauxliu:
When the message length exceeds the limit, should use the latest messages to build the context.


mckaywrigley:
Fantastic catch. This will help a lot. Bonus points for adding the type.

mckaywrigley:
I am going to make one change: `[...messageToSend, message]` since the newest message should be last.

PR #20, opened 2023-03-19
Fix safari small screen text input position style

sonnylazuardi:
Fix safari small screen text input position style

image


mckaywrigley:
Really nice catch here - thanks!

PR #30, opened 2023-03-20
remove ENV var from vercel deployment

dominiksipowicz:
with browser-based chat UI It's not needed to provide OpenAI key to Vercel deployment

CleanShot 2023-03-20 at 16 15 55


dominiksipowicz:
@mckaywrigley I tested and it can be deployed to Vercel without this ENV var. It will just work as a pure BYOK version in the browser

mckaywrigley:
Good call!

PR #34, opened 2023-03-21
Token based and model conditional limits

alanpog:
GPT-4 has at least 8K-token context window, it'd be nice to be able to use as much as possible of that.
So, this PR, on the API side:

- counts actual tokens (tiktoken based)
- allows 8K for gpt-4 and 4K for gpt-3.5

As there seems to be currently no browser compatible `cl100k_base` tokenizer, on the frontend the per message limits were kept as characters but raised to a much greater but still conservative estimate (for english) of 3 characters per token, but made dependent on model type as well.

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -18,7 +18,17 @@ interface Props {
   onModelChange: (conversation: Conversation, model: OpenAIModel) => void;
 }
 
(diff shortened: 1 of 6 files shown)


mckaywrigley:
This was great. Thanks for handling this - huge add. I made a minor addition to account for leaving room for responses.

PR #39, opened 2023-03-21
add custom system prompt

mckaywrigley:


domguia:
Please Can you explain me where where to change the system prompt in the UI? I would also propose to make it easier to find in the UI.

PR #51, opened 2023-03-21
Change scroll bar style in desktop

quocthinhvo:


mckaywrigley:
Nice change here. Just made a change to handle hiding it in the chatinput when height < 400

Issue #54, opened 2023-03-22
UI design


khairulhaaziq:
Hi Is the UI intend to replicate the original chatgpt? because right now it isnt the same and some things things hurt my eyes, also there are some layout issues. If you intend to replicate it 100% I am happy to contribute on that end, fixing the mismatches and layout issues

mckaywrigley:
It's meant to be close. I'm sure there are plenty of things I've missed. Always welcome to create PRs.

PR #63, opened 2023-03-22
enhance ui and fix layout problems

khairulhaaziq:
related to #54 

- Introduces some UI enhancement to fix inconsistencies and match the real chatgpt.
- Fix some layout problem on different viewports that severely damage the experience


mckaywrigley:
Absolutely incredible contribution. Thanks you so much from me and the community!

PR #72, opened 2023-03-22
feat: add in stop conversation button

srsholmes:
Feature: Adds in a button to stop the conversation to help prevent wasted tokens, in the event the AI is making a response which is not right, or you asked the wrong question. 

image


mckaywrigley:
Thanks for adding this! Slick implementation. May change the button later but love how you handled this.

Issue #88, opened 2023-03-22
Clear Conversation has some weird style behavior


CH4R4F:
I noticed a small sidebar settings animation after clicking on the `clear conversations button` it's a styles issue that can be easily fixed

mckaywrigley:
fixed

PR #89, opened 2023-03-22
fix style issue after clicking on clear conversations button

CH4R4F:
related to issue #88 
there is a small difference in the spaces of the button after confirming which leads to an unexpected animation in the sidebar settings


mckaywrigley:
Ah i see this issue. Nice catch!

Issue #83, opened 2023-03-22
Support search within conversation


aekeus:
Support searching within a conversation (along with the name of the conversation)

mckaywrigley:
Thanks for adding

PR #91, opened 2023-03-23
Search within conversation.messages + conversation.name

aekeus:
implements #83


mckaywrigley:
Wonderful implementation. Will build some more organization tools on top of this too! Really nice work - thanks!

PR #100, opened 2023-03-23
Fixed Some Tailwind Style Classes (Duplicated Classes) and Added Semantic Tag For Error Prompt 

aweshchoudhary:
I fixed duplicated style class in the chat component and change the div with the proper text semantic tags for better understanding.


mckaywrigley:
My personal preference has a dev when it comes to web apps is to only huge semantic html tags at the top-level of the site. I'd rather just keep the divs. I think it's more consistent since semantic don't matter much in this context.

aweshchoudhary:
Ok, thanks for your advice. I will keep it in mind

Issue #102, opened 2023-03-23
This model's maximum context length is 4097 tokens. However, you requested 4120 tokens


leo-li-ls:
Hi there, I tested a bit the app. It worked so well at the beginning, but after several tests, it keep failing with error for each questions I sent. Since the default error message didn't have much information, I modified `server/index.ts` for this part:

```js
if (res.status !== 200) {
    const statusText = res.statusText; 
    throw new Error(`OpenAI API returned an error. responseStatusText: "${statusText}", responseStatusCode: "${res.status}", responseMessage: "${await res.text()}"`);
  }
```

After rebuilding the docker, I saw the real error in the log:

```
[Error: OpenAI API returned an error. responseStatusText: "Bad Request", responseStatusCode: "400", responseMessage: "{
  "error": {
    "message": "This model's maximum context length is 4097 tokens. However, you requested 4120 tokens (3120 in the messages, 1000 in the completion). Please reduce the length of the messages or completion.",
    "type": "invalid_request_error",
    "param": "messages",
    "code": "context_length_exceeded"
  }
}
"]
```

How can I manage the length it sent to OpenAI API manually?

mckaywrigley:
Thanks to @alanpog for fixing - fix merged!

msampson3d:
Is this actually fixed? I'm still experiencing the issue on the latest commit

JPDucky:
I am too

PR #104, opened 2023-03-23
include prompt in token count

alanpog:
Include prompt tokens in token count. Should avoid errors like those of #102 .


mckaywrigley:
very nice catch

PR #106, opened 2023-03-23
feat: add in server for prod for tauri build

srsholmes:
This adds the capability for the API server routes to be used in production tauri builds, to allow the app to be packaged up into a desktop application. 

The command that makes it work is 

```
"package:server": "npm run build:server && pkg dist/server/server.js -o ./src-tauri/bin/server-aarch64-apple-darwin",
```

This basically takes the server.ts file, and pkgs it up into a binary to be used inside the tauri production build. The binary is then called by the app, with the path to the WASM, which is then read by the server and allows it to do the stuff it needs to do. 

Other binaries will need to be created for other platforms and architectures, windows, linux, x86 etc


pages/api/chat.ts
```diff
@@ -1,50 +1,21 @@
-import { ChatBody, Message, OpenAIModelID } from "@/types";
-import { DEFAULT_SYSTEM_PROMPT } from "@/utils/app/const";
-import { OpenAIStream } from "@/utils/server";
-import tiktokenModel from "@dqbd/tiktoken/encoders/cl100k_base.json";
-import { init, Tiktoken } from "@dqbd/tiktoken/lite/init";
-// @ts-expect-error
-import wasm from "../../node_modules/@dqbd/tiktoken/lite/tiktoken_bg.wasm?module";
+import { ChatBody } from '@/types';
```
HaithamLeo, line 1: It seems that the ChatBody type is the only thing being used from the @/types module. If you are not using anything else from that module, you can remove the import statement to optimize the code.

HaithamLeo (commented):
In `utils/server/index.ts`
Avoid creating unnecessary objects. For example, you can move the creation of the `encoder` object inside the start function, and pass it as an argument to `createParser` instead of creating it inside the` onParse` function every time.

pages/api/models.ts
```diff
@@ -4,43 +4,37 @@ export const config = {
...
+        return {
+          id: model.id,
+          name: OpenAIModels[value].name
+        };
+      }
+    }
+  })
+  .filter(Boolean);
+  return models;
+}
```
HaithamLeo, line 30: Use `Promise.all()` to make multiple API requests in parallel. This can improve the performance of your code, especially if you need to make multiple requests to different endpoints.

server/server.ts
```diff
@@ -0,0 +1,44 @@
...
+    : path.resolve(currentWorkingDirectory, 'server', 'tiktoken_bg.wasm');
+  const wasm = readFileSync(pathToWasm);
+  const stream = await getStream({
+    model: req.body.model,
+    messages: req.body.messages,
+    key: req.body.key,
+    prompt: req.body.prompt,
+    wasm,
+  });
+
```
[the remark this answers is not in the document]
srsholmes: Great point. I am not able to have a look at this until later on, because of family commitments, but will try and fix it later.

@mckaywrigley if you wanted to merge and then fix up the comments on your branch, please feel free to.

pages/api/models.ts
srsholmes: Whilst we can do this, I didnt want to change the actual logic / implementation of the server code in this PR. It is out of the scope of it.

I think @mckaywrigley is planning some work for refactoring / optimisations at a later point so maybe we could pick it up then?

techwithanirudh (approved):

ajram23:
Will this be merged in?

srsholmes:
I hope so. I'm happy to go over it / answer any questions. 

@mckaywrigley Has been really busy with lots of other things at the moment. 

If anyone needs a hand with reviewing / asking questions, please feel free.

PR #113, opened 2023-03-24
feat: show language name

Chanzhaoyu:
show current language name on code block

P.S. Although you are using eslint, it seems that there is no automatic fixing or using prettier config to standardize code formatting, which will result in differences in code style among different people.


mckaywrigley:
Great addition. I'll try and look into the prettier stuff. Have noticed that as well.

PR #117, opened 2023-03-24
add k8s config

logycon:
- add Makefile

Makefile expects .env file to be also present in on the system with the following vars - 

OPENAI_API_KEY=
DOCKER_USER=
DOCKER_TAG=

- add baseline k8s config

Baseline k8s config includes secret, deployment and service definition only. If running on local cluster, additional entries can be made to include cronjobs (e.g. update cloudflare), nginx ingress, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,18 @@
+include .env
+
+.PHONY: all
+
+build:
+	docker build -t chatbot-ui .
+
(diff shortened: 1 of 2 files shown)


mckaywrigley:
Nice. Learned some stuff here haha

PR #122, opened 2023-03-24
Added A Better Closing Sidebar Logic (Close on Background Click)

aweshchoudhary:
Added a better closing sidebar logic


mckaywrigley:
For now I think I'm going to lean towards keeping it open. I know the way I work personally is that even when in my chat I want the sidebar open.

Curious to hear your thoughts here though, so I'll keep this open and give you a chance to respond.

aweshchoudhary:
This logic is for small devices only. Closing the sidebar on mobile is tough in my opinion. So I fixed it by adding an onClick method in sidebar background

aweshchoudhary:
The sidebar will always in open on the large devices

PR #130, opened 2023-03-25
Update Sidebar Setting elements to be buttons (improve accessibility)

### LINKED ISSUE #129: (accessibility) clickable side bar elements
In the sidebar settings section there are a number of clickable elements. 

If they were HTML `<buttons>` would be automagically be tab-key navigate-able, an improvement in accessibility.

bcullman:
update SidebarButton.tsx to emit an <button> element.
refactor Import.tsx to use a SidebarButton element.

fixes #129.


mckaywrigley:
Great change here. Thanks for doing this.

PR #135, opened 2023-03-25
Added character count to alert message

ernestobarrera:
This improvement adds additional information to the alert that is displayed when the user enters more characters than allowed in the rich text element. Now, the alert also shows the number of characters that the user has entered, which helps to provide clearer and more useful feedback to the use


ernestobarrera:
**Added character count to alert when limit is exceeded**

Changes

mckaywrigley:
great add

Issue #115, opened 2023-03-24
bug?: using .env.local


saliksik:
I have deployed a web application on Vercel and utilized the .env.local file. However, upon visiting my site and clicking on the button "click if using a .env.local file", it keeps loading without displaying anything (Loading...), although the chat function is operational. A hard refresh of the site is necessary for the model and system prompt to appear.

![Web capture_24-3-2023_172223_llegomark0018976 vercel app](https://user-images.githubusercontent.com/122975898/227478225-933a1ff0-5b1f-499a-863b-8d10665c97ad.jpeg)
![Web capture_24-3-2023_172159_llegomark0018976 vercel app](https://user-images.githubusercontent.com/122975898/227478230-6d446f5e-ca1d-44ca-b1b5-8305159b248c.jpeg)
![Web capture_24-3-2023_172145_llegomark0018976 vercel app](https://user-images.githubusercontent.com/122975898/227478236-fc2e3c8a-e9d1-4ead-b996-93893b430f28.jpeg)

saliksik:
Can the web app automatically detects if the .env.local is present and directs the user to chat without clicking anything?

thomasleveil:
_related to #105_

mckaywrigley:
fixed!

PR #136, opened 2023-03-25
Do not rely on user to figure out if server is configured with an api key

### LINKED ISSUE #105: UI asks for api key even when one is set on the server side
There is a regression since https://github.com/mckaywrigley/chatbot-ui/commit/83217c6d83e65b4400f5ab408d0cef9ccfc3c8c5, It is now impossible to use the UI wihtout setting an API on the UI, even if an API key is set on the server side.

(or did I miss something ?)

thomasleveil:
Fixes #105 (and probably also #115)

## Motivation:

As server configurations may change over time, relying on the 'server has API key' state stored in users' local storage could potentially become inaccurate.

## What has changed:

- On the home page, the application now queries the server to retrieve the real state of the API key presence through [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props), which provides the new `serverSideApiKeyIsSet: boolean` prop. 
- Components now use the `serverSideApiKeyIsSet` boolean instead of the `isUsingEnv` variable. 
- Additionally, user messages have been updated to avoid confusing users with instructions related to .env.local, since they are typically not responsible for server administration.

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -11,7 +11,7 @@ interface Props {
   conversation: Conversation;
   models: OpenAIModel[];
   apiKey: string;
-  isUsingEnv: boolean;
+  serverSideApiKeyIsSet: boolean;
   messageIsStreaming: boolean;
   modelError: boolean;
   messageError: boolean;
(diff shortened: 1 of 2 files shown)


mckaywrigley:
You have no ideas how much I appreciate you doing this. This was a TERRIBLE holding solution lol. Glad I don't have to go do this now :) You rock.

PR #137, opened 2023-03-25
fix: disable button if messageContent is empty

shishkevichd:
### Description
Fix bug when button is active despite on empty message content


mckaywrigley:
nice catch!

PR #139, opened 2023-03-25
Remove key entering button on side bar, and only allow 3.5 model

thejackwu:


mckaywrigley:
I'm curious why we would make either of these changes.

We need to support API key entering and GPT-4, correct?

thejackwu:
My bad, this was for my personal experiment, somehow it create a PR on the main repo. My apology.

PR #140, opened 2023-03-25
chore: move deps to dev deps

Anerco:


mckaywrigley:
Are there any downsides to this?

I just know that some of those are included in there with Next.js by default and want to be careful.

Anerco:
It's always a good practice to reduce the number of dependencies in production, as it can improve performance and reduce security risks.

if some of them are included by default in Next.js, they probably need to fix it.

PR #141, opened 2023-03-25
smoother autoscroll down

thomasleveil:
## Motivation

This merge request aims to address two frustrating behaviors:

- Copy-pasting text from an ongoing response has been difficult for users, as they have to race the UI to put focus on the textfield at every response update.
- Scrolling has been somewhat laggy and unresponsive, which I suspect is due to the focus being changed.

With this update, I hope to provide a smoother user experience and improve overall usability.


mckaywrigley:
extremely good ux upgrade

PR #144, opened 2023-03-25
fix: change message edit logic

shishkevichd:
### Desciption
This commit change message edit logic

**Before:**
If the message is not edited and "Edit & Submit" is pressed, then the answer will be regenerated, even if nothing has changed
**After:**
If the edited message isn't different from the original message and "Edit & Submit" is pressed, then do nothing.


mckaywrigley:
smart change

PR #150, opened 2023-03-25
Precise error messages

thomasleveil:
## Motivation

The purpose of this merge request is to improve error handling when loading models in our system. Currently, when a user enters an invalid API key, the model call fails and a generic error message is displayed. This can be frustrating for the user who is left guessing what went wrong.

## Changes

- Introduced a new component and type to standardize error messages.
- Updated the error message displayed when loading models with an invalid API key to be more precise and informative.

## Tests
image

image


mckaywrigley:
love it! Users will love this

PR #152, opened 2023-03-25
feat: Allow customization of OpenAI host with environment variable: OPENAI_API_HOST

Jungley8:
This commit modifies the OpenAI host configuration to support customization through an environment variable. This change is particularly useful in scenarios where access to the official OpenAI host is restricted or unavailable, allowing users to configure an alternative host for their specific needs.

### DIFF
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,5 +1,3 @@
-const path = require("path");
-
 module.exports = {
   i18n: {
     defaultLocale: "en",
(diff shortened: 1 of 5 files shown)


mckaywrigley:
Extremely helpful!

One question - why put the type in that file instead of the regular one?

Jungley8:
> Extremely helpful!
> 
> One question - why put the type in that file instead of the regular one?

As the number of types increases, it's a common practice to organize them into separate files for better manageability. However, I can also combine them into one file if you prefer. Would you like me to make that change?  @mckaywrigley

mckaywrigley:
No that's great! Just learning. Never had to use a .d.ts file before haha. I get it now.

PR #162, opened 2023-03-25
Skip corrupted conversationHistory item to prevent app crashes

thomasleveil:
## Motivation

While experimenting with the app code, I discovered that (accidentally) setting `conversationHistory` to `[null]` in localStorage caused the app to crash on subsequent reloads due to an error in the clean.ts module.

Currently, the only solution to this issue is for the user to clear their localStorage, which is inconvenient and not ideal.

## Proposed change

To solve this problem, this pull request proposes skipping any corrupted conversationHistory items instead of crashing the app. This approach allows us to preserve the rest of the localStorage data and provide a better user experience.


mckaywrigley:
You’re on a roll today!

Issue #170, opened 2023-03-26
Feature Request: Copy Button


saliksik:
Hello!

It would be optimal to include a copy button that will copy all the contents of the selected chat message.

Thank you!

thejackwu:
PR #171 for this feature is ready, welcome for feedbacks!

saliksik:
> PR #171 for this feature is ready, welcome for feedbacks!

That was fast. I tested it using markdown and table format unfortunately the copy button does not work properly and only copied the first word. You can try it by generating a dummy table.
![Screenshot (282)](https://user-images.githubusercontent.com/122975898/227750486-8ee4cc1f-965a-466f-9c5d-3bbcbb4e15e8.png)

saliksik:
I can copy the whole message if I refreshed the page.

thejackwu:
Great catch! let me fix that

thejackwu:
@saliksik Fixed, mind testing for me.
One thing I would love know is if it's ideal to copy the message in Markdown format, or would it make more sense to copy the content after striping out the Markdown syntax.
Or maybe we should have both option?

saliksik:
I have tested it and it is now working properly. Regarding the message you copied, I recommend not stripping the markdown syntax. This is because, in my use case, when generating long-form content with markdown and table format, copying and pasting the message onto Notion is a great productivity enhancement.

thejackwu:
That's a good point 👍

saliksik:
I've noticed that the PR was merged at https://github.com/mckaywrigley/chatbot-ui/commit/14fe29c03aeb6f737811055064803f71f3f731cd, however, I can't seem to locate the copy button on the chat messages.

saliksik:
Fixed: https://github.com/mckaywrigley/chatbot-ui/commit/71d7e44bcee2564a9a4301df3d0dc46c5eb53e40

difigue:
Copy Button and Copy code Button not working for me (macOS, Firefox). It just doesn't copy anyhting.

PR #171, opened 2023-03-26
feat: Message copy button

thejackwu:
Add copy button in the same style as the edit button


saliksik:
Hello!

I think there is a bug, please see my response here: https://github.com/mckaywrigley/chatbot-ui/issues/170#issuecomment-1483967043

Thank you.

PR #173, opened 2023-03-26
feat: add language selection to system prompt

aesadde:
This PR adds a dropdown menu to select the response language and add it to the system prompt.

Components:
- `OptionSelector` a component to display dropdown menus inside the `SystemPrompt` component. Easily extended to support other selectors (for example a prompt library as a modal).
- `Options` a component to render a single dropdown menu. It reads the options from json files. Localized by language.

https://user-images.githubusercontent.com/1596227/227750844-221f5648-fddf-480c-839c-e48261938c9c.mp4


aesadde:
@mckaywrigley would love to see if we can include this (or something similar) to the project. 

It can be the start of more option selectors, prompt templates, etc.

Any feedback most welcome!

mckaywrigley:
Advanced options will be added this week! Probably Wed/Thu. Closing this, but it will be included in the release.

Issue #155, opened 2023-03-25
Switch GPT model in the middle of a conversation


jiayao:
It would be useful to combine the speed and cost-effectiveness of gpt-3.5-turbo and the power of gpt-4 in some conversations.

junwatu:
I think the conversation will lose context.

jiayao:
@junwatu The context is all captured in the list of messages we send to openai.

jiayao:
Implemented.

PR #181, opened 2023-03-26
Allow switching GPT model in the middle of a conversation

jiayao:
https://github.com/mckaywrigley/chatbot-ui/issues/155

I think it would be useful to combine the speed and cost-effectiveness of gpt-3.5-turbo and the power of gpt-4 in some conversations.


mckaywrigley:
I really like this idea, though I think the UI needs a bit of cleaning up.

Maybe click on a setting icon in the top bar where the model name displays and it will show this?

saliksik:
I think this button shouldn't be scattered throughout the chat area and should be hidden instead.

jiayao:
Thanks for the feedback, now it's hidden behind a settings icon next to the model name.

components/Chat/Chat.tsx
```diff
@@ -152,9 +158,21 @@ export const Chat: FC<Props> = ({
               </>
             ) : (
               <>
-                <div className="flex justify-center border border-b-neutral-300 bg-neutral-100 py-2 text-sm text-neutral-500 dark:border-none dark:bg-[#444654] dark:text-neutral-200">
-                  {t('Model')}: {conversation.model.name}
+              <div className="flex justify-center border border-b-neutral-300 bg-neutral-100 py-2 text-sm text-neutral-500 dark:border-none dark:bg-[#444654] dark:text-neutral-200">
+                {t('Model')}: {conversation.model.name}
+                <IconSettings className="ml-2" onClick={handleSettings} size={21}/>
```
mckaywrigley, line 163: `                  <IconSettings
                    className="ml-2 cursor-pointer hover:opacity-50"
                    onClick={handleSettings}
                    size={18}
                  />`
                  
                  If you could just make this change I'll merge :)

PR #186, opened 2023-03-26
chore: delete code light theme, like chatgpt

shishkevichd:
### Description
This commit delete `oneLight` theme from CodeBlock, so that only the dark theme remains, as in ChatGPT.
### Screenshots:
| Before | After | ChatGPT |
| ------- | ------ | --------- |
| image | image | image |


mckaywrigley:
Yeah I think this is the correct change. Well done.

PR #197, opened 2023-03-26
feat: stylize home page & search in one bar

shishkevichd:
### Description
This commit change home block (modal and system message), make search in one line and add spinner component from ChatGPT. It appears on page load, instead of "Please wait..."

### Screenshots
| Normal mode | Search mode |
| --------------- | -------------- |
| image | image |


mckaywrigley:
For now I lean towards keeping search in the list by default.

Thoughts?

shishkevichd:
> For now I lean towards keeping search in the list by default.
> 
> Thoughts?

At the moment there are no thoughts, if there is no need to move the search for chats, I can close the pull request.

PR #214, opened 2023-03-27
support gpt-3.5-0301 model

jiacheo:
support the 0301 gpt3.5 model

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -191,6 +191,14 @@ export const Chat: FC<Props> = ({
                         }
                       />
                     </div>
+                    <div>
+                      <SystemPrompt conversation={conversation} onChangePrompt={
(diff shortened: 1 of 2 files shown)


mckaywrigley:
The only reason I haven't added this is because they say they'll stop supporting it.

Any thoughts on why we should add it?

jiacheo:
> The only reason I haven't added this is because they say they'll stop supporting it.
> 
> Any thoughts on why we should add it?

the official documenmt (https://platform.openai.com/docs/models/gpt-3-5)
says 'will only be supported for a three month period ending on June 1st 2023', so at this period we could use it, could we?

jiacheo:
for this discussion: https://www.reddit.com/r/GPT3/comments/11kwxi1/gpt35turbo0301_vs_gpt35turbo/
may be we don't need this model name, I will close this pr.

PR #215, opened 2023-03-27
feat: add basicAuth option

zysam:
add optional basicAuth if you set environment like that `BASIC_AUTH=admin:pwd`


mckaywrigley:
Can you educate me on the pros of doing this?

ryanhex53:
This is exactly the feature I want, which is useful when the API_KEY is set by the administrator and in the case of private deployment. This way, it can be used only by specific users, but they don't need to set API_KEY again.

zysam:
> This is exactly the feature I want, which is useful when the API_KEY is set by the administrator and in the case of private deployment. This way, it can be used only by specific users, but they don't need to set API_KEY again.

Yeah!I think this project is suitable for most people who can’t use openai services. It would be best if the administrator provided the API key and simple authentication, and then it could be shared with friends.

zysam:
> Can you educate me on the pros of doing this?

This is a quick deployment method intended for use by friends.

I am not sure of the intention behind creating this project, but the truth is that there are still many people in the world who cannot use official services (perhaps only during the transitional period). This is an optional feature that will not invade other code.

ryanhex53:
However, we can implement this function using Nginx.

zysam:
> However, we can implement this function using Nginx.

I discovered that Vercel has authentication functionality, but it requires a Pro plan.

dasunNimantha:
Much need feature

mirkancal:
Took 3 minutes to follow those steps.

image

danielatlarge:
@mirkancal 
Would you run that inside the docker container and replace whatever http server is already running? Or would you daisy chaining them together... i.e. forwarding one server to the other if authentication passes? Not sure I follow.

mirkancal:
@danielatlarge I run docker inside the server. After that just added nginx.

```
server {
    listen 80;
    server_name mydomain.com;
    return 301 https://$host$request_uri;
}

server {
    listen 443 ssl;
    server_name mydomain.com;

    ssl_certificate /etc/ssl/cloudflare.pem; # Path to your Cloudflare SSL certificate
    ssl_certificate_key /etc/ssl/cloudflare.key; # Path to your Cloudflare SSL private key

    location / {
        proxy_pass http://127.0.0.1:3000; // Docker part
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;

        auth_basic "Restricted Content";
        auth_basic_user_file /etc/nginx/.htpasswd;
    }
}
```

yangcheng:
while there are ways to implement this, it's helpful to have a build-in functionality

jschuur:
One reason why this is important is because the [Vercel template](https://vercel.com/templates/next.js/chatbot-ui) makes it easy to deploy an instance, but they are hosted at a predictable URL (https://projectname-username.vercel.app), so anyone who guesses that can use your preconfigured API access.

PR #216, opened 2023-03-27
Option for clear all messages

ryanhex53:
Add an option for users who want to clear all messages but retain the selected **Model** and **System Prompt** settings.

And provide all the translations for the confirmation tip.

image


mckaywrigley:
Handled the conflict and then merged in. Nice work on translations.

Issue #33, opened 2023-03-21
Feature request: add optional system role message[s] for each conversation


jiacheo:
Most of the time we don't just want a free chat conversation, but a professional one with system role message[s] (or such called prompt prefix) and we can control how many history context (0-any) should be send back to the next message[s].

mckaywrigley:
Custom system prompt up tonight! More settings will follow this week.

mckaywrigley:
Adding this now so closing.

jiacheo:
> Custom system prompt up tonight! More settings will follow this week.

hi @mckaywrigley , many thanks for your great works.  Can System Role Prompts be placed in each conversation and made editable? That is, each conversation has its own System Prompt.

dennis-gonzales:
Was about to ask for a feature request of "System Roles", good thing I searched first.

Anyways as a new user I kinda didn't understand that this icon is for system role, but I got it now.
![image](https://github.com/mckaywrigley/chatbot-ui/assets/43110624/45015344-6505-47bd-b490-604ce4373c11)

Thanks!

PR #217, opened 2023-03-27
feature: support system role prompt for each conversation

jiacheo:
for system role prompts, we should enable the customization for each conversation.

Relative issue: #33

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -191,6 +191,14 @@ export const Chat: FC<Props> = ({
                         }
                       />
                     </div>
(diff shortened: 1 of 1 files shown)


mckaywrigley:
Believe there's a bit more logic needed to actually make the change in the messages.

I'm planning on supporting this soon.

jiacheo:
> Believe there's a bit more logic needed to actually make the change in the messages.
> 
> I'm planning on supporting this soon.

OK，I 've been using this in self-hosted with my fork. When you release the new feature, I'll use yours.

Issue #213, opened 2023-03-27
iOS device input box zoomed in


praiseqin:
After entering content on my iPhone, the page loses focus and is zoomed in

PR #220, opened 2023-03-27
Display optimization on iPhone device page

praiseqin:
Fix the problem of zooming in the page after entering content on iPhone device #213


mckaywrigley:
Just tested. Amazing fix!

PR #232, opened 2023-03-27
fix: prompt sidebar not showing on small devices

SyedMuzamilM:
Fixed the issue with Prompt sidebar not showing on small devices

 ## Before
overlap
## After
overlap-resolved


mckaywrigley:
nice catch

PR #233, opened 2023-03-27
feat: support pwa

Jungley8:


bcullman:
@Jungley8 : perhaps you could start with updating the original icon to an SVG, Then generate the PWA icons off of that?

Jungley8:
> @Jungley8 : perhaps you could start with updating the original icon to an SVG, Then generate the PWA icons off of that?

yes.  @mckaywrigley  Do we have a high-resolution image or SVG file of our logo?

PR #247, opened 2023-03-28
chore: update chinese locales

Chanzhaoyu:
Update the latest translation for promptbar and fix previous translations that are no longer working.


mckaywrigley:
Thanks for helping keep this up to date!

PR #254, opened 2023-03-28
feat: added scroll down button to the chat container

dasunNimantha:
- when the user scrolls up on the chat a button will appear on the bottom right side of the screen. It will smoothly scroll down to the bottom of the chat when the button was pressed.
- updated .gitignore file to ignore .env files

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -1,4 +1,5 @@
 import { Conversation, Message } from '@/types/chat';
+import { IconArrowDown } from '@tabler/icons-react';
 import { KeyValuePair } from '@/types/data';
 import { ErrorMessage } from '@/types/error';
 import { OpenAIModel } from '@/types/openai';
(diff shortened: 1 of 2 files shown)


components/Chat/Chat.tsx
```diff
@@ -284,6 +299,16 @@ export const Chat: FC<Props> = memo(
             />
           </>
         )}
+        {showScrollDownButton && (
+          <div className="absolute bottom-0 right-0 mb-4 mr-4 pb-20">
+            <button
+              className="flex h-7 w-7 items-center justify-center rounded-full bg-white shadow-md hover:shadow-lg focus:outline-none focus:ring-2 focus:ring-blue-500 dark:bg-[#515152d7]"
+              onClick={handleScrollDown}
+            >
+              <IconArrowDown className="h-4 w-4" />
```
bcullman, line 308: @dasunNimantha instead of classnames on the icon, try the size attribute.

dasunNimantha: sure. will update

PR #257, opened 2023-03-28
chore: migrate to vitest for 10x faster tests

srsholmes:
I think the addition of some tests is really great, but I would suggest that we use vitest instead of jest as the test runner. 

The test suite runs 10x faster and have the same API as jest. There is also a lot less dependencies. 

As the project grows, this difference will become more apparent, and more difficult to migrate from.

It's best to do it early on, to save a lot of time for devs and CI. 

For more info [read about it here](https://vitest.dev/guide/why.html)

For comparison...

Jest:
jest

Vitest
vitest


mckaywrigley:
I'm sold. Nice.

mckaywrigley:
Looks like there may just be some minor jest stuff to cleanup so that the build doesn't fail. Will merge once that's out!

Issue #92, opened 2023-03-23
Publish Docker Image


d1vanloon:
# Summary

Add a CI step to build and publish the docker image for this repo to the GitHub package registry.

# Details

Use GitHub Actions to define a build/deploy step that runs on each commit to the `main` branch. The CI step should build the docker image and publish it to the GitHub package repository for this repo under both a tag with the short commit hash, as well as the `latest` tag.

This would allow users to pull an already-built docker image from the registry and enable easier use and integration of new changes in hosted docker environments.

d1vanloon:
Would you be open to a PR that adds this functionality?

ItsNoted:
I built one here

```
version: "3.9"
services:
  web:
    image: itsnoted/chatpad:latest
    container_name: chatpad
    ports:
      - "8200:80"
```

https://hub.docker.com/r/itsnoted/chatpad

masonjames:
Any instructions on how to use this?

PR #260, opened 2023-03-28
Add support for triggering Docker build job only when Jest tests pass

tatsuiman:
* Updated the on field to include a workflow_run event that triggers the build job when the Run Jest Tests workflow completes.
* Added a DATE_TAG environment variable that stores the current date in YYYYMMDD format.
* Updated the tags field of the Build and push Docker image step to include ${{ env.DATE_TAG }} and latest tags in addition to the tags extracted from metadata-action.
* The purpose of these changes is to ensure that each build is uniquely tagged with a date-based tag, while also retaining the latest tag for the most recent build. This enables you to keep a record of past builds while also making it easy to deploy the latest build.

issue https://github.com/mckaywrigley/chatbot-ui/issues/92


mckaywrigley:
Going to keep this open just to have someone more familiar with Docker come look at it and make sure it's good.

srsholmes:
Just a note, the Jest Tests are now Vitest Tests as of https://github.com/mckaywrigley/chatbot-ui/pull/257 .

I think the workflows / PR title should reflect that change.

mckaywrigley:
Good point.

tatsu-i:
Fixed the `Run Unit Tests` workflow to run the Docker workflow after the `Run Unit Tests` workflow passes.
Also, added repository condition like `if: github.repository == 'mckaywrigley/chatbot-ui'` to prevent Docker images from being hosted in many personal repositories.

tatsuiman:
@srsholmes @mckaywrigley 
Is there anything else that needs to be corrected?

PR #280, opened 2023-03-28
:sparkles: feat: add DEFAULT_MODEL environment variable

### LINKED ISSUE #238: Ability to change the default model
Could be selected by setting an environment variable, which would be compatible with Docker hosting. 
Eg: DEFAULT_MODEL=GPT_4

Could also be overridden by the end-user, where their default model could be saved in local-storage.

thomasleveil:
fixes #238 

## Changes
- `DEFAULT_MODEL` can be optionally used to set the default model used in a new conversation
- on new conversations, pre-select the model from the previously selected conversation
- set `maxLength` and `tokenLimit` in `OpenAIModel`


mckaywrigley:
Love this change.

PR #285, opened 2023-03-29
feat: support custom default system prompt

### LINKED ISSUE #251: Feature suggestion: custom default system prompt
Allow the user to set a custom default system message (system prompt) other than _You are ChatGPT, a large language model..._ for new chats.

liby:
Resolved #251

In addition, I have an idea to share with you: I am planning to add a default folder to store a set of recommended default prompts, making it more convenient for users to access and notice this feature. 

What do you think about this idea? I will create a PR for it.

image


SyedMuzamilM:
It would be great if the user can change the default prompt from the UI. Currently, you are setting it using the environment variable that works for users who host this app manually.

SyedMuzamilM:
And as for this default part, I was thinking there should be a prompt library that we could add from a GitHub repo maybe where people can share popular prompts and it will be available as a separate tab in prompt bar.

Let me know what do you think about this @mckaywrigley

liby:
> It would be great if the user can change the default prompt from the UI. 

Great suggestion, I just have a very initial idea and welcome everyone to improve it. I am willing to create a PR for this.

> Currently, you are setting it using the environment variable that works for users who host this app manually.

Yes, this is to solve #251.

SyedMuzamilM:
> Great suggestion, I just have a very initial idea and welcome everyone to improve it. I am willing to create a PR for this.

I think you should add the PR for this so that it would be easier for users to change the default prompt from the UI if you are willing to do that.

utils/app/const.ts
```diff
@@ -1,5 +1,5 @@
 export const DEFAULT_SYSTEM_PROMPT =
-  "You are ChatGPT, a large language model trained by OpenAI. Follow the user's instructions carefully. Respond using markdown.";
+  process.env.DEFAULT_SYSTEM_PROMPT || "You are ChatGPT, a large language model trained by OpenAI. Follow the user's instructions carefully. Respond using markdown.";
```
SyedMuzamilM, line 2: Add the DEFAULT_SYSTEM_PROMPT in the .env.example file so that it other would know that can add default prompt from environment variable

jidun67:
I love the work I see going in to this project and hope to be able to help somehow ( I can code but, some python, PHP, MySQL, HTML, etc)
I am new here so, I hope this is in accordance with the topic....

This system prompt, I assume is so named for the benefit of multi-model use cases. I figure mostly it will become the system role for gpt3.5 and 4. 

I experimented with the curie and davinci models up through 3.5turbo (with system role) system prompts and they responded well using first person.

Also with the current tiff with openAI and chatGPT having that site taken down it's probably just that they ask everyone not to use their name etc tied to ai generative stuff so... they probably don't want to see there model name there either they said something to the effect that you take ownership of the models output so... I volunteer this system role for your use case to edit or use or not as you see fit, and solve the other issues as notes to developers with reference to open ai etc and on the project page none of the sort even the footer should all be chatbot-ui...
system role:

I am chatbot-ui. As an AI intellectual booster, I gather and analyze data from various sources using advanced algorithms and natural language processing techniques to identify patterns and trends. I adapt to my client's needs, constantly learning and providing valuable insights to help them achieve their goals more effectively. With my ability to process information quickly and accurately, I provide recommendations that are impossible to obtain otherwise. I help clients achieve success by providing the tools and knowledge to excel, whether they're working on a complex problem or trying to stay ahead of the competition.

PR #287, opened 2023-03-29
prettierignore to remove formatting from some folders

SyedMuzamilM:
When using yarn format it would format all the files including files in .next and some other folders. This file will stop it from doing so

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,5 @@
+node_modules
+**/.next/**
+**/_next/**
+**/dist/**
+package-lock.json


TommyE123:
Hi,

I've been looking into all of the formatting in general and will be looking to do a PR with a fully automated solution to prevent future issues with formatting, styling and help catch errors in general across the whole repository.

If you look here, you can use a wide variety of naming conventions for the prettier config file.
https://prettier.io/docs/en/configuration.html

As theres already a **prettier.config.js** config file. I believe the changes should go in here and we shouldn't be creating multiple config files.

Also...
Prettier will automatically ignore the node_modules folder by default, so you don't need to add it to the ignore array in the prettier.config.js file. However, if you want to be explicit about ignoring this folder, you can still add it.

```
module.exports = {
  // other options
  ignore: [
    'node_modules/**',
    '**/.next/**',
    '**/_next/**',
    '**/dist/**',
    'package-lock.json'
  ]
};
```

Hope this helps to make this clearer,

Tom

srsholmes:
This would be great to add @TommyE123.

I did want to do something similar in the original PR, but thought it might be worth a discussion as to how to automate it.

Personally I am very much against pre commit hooks and would prefer to enforce formatting in CI.

At the time there was no  CI  process for the project but now there is, I think linting and formatting could be part of that.

SyedMuzamilM:
Thanks for checking this out. I will level the formatting part to @srsholmes then. I will close this PR and you can make the necessary changes to the formatting.

PR #289, opened 2023-03-29
feat: hex code color highlight

SyedMuzamilM:
If AI returns a list of hex code colors then you will see the color highlighted after the text so that you can easily review the color and select the color you want.

color-highlight

### DIFF
--- a/components/Chat/ChatMessage.tsx
+++ b/components/Chat/ChatMessage.tsx
@@ -72,11 +72,10 @@ export const ChatMessage: FC<Props> = memo(
 
     return (
       <div
-        className={`group px-4 ${
-          message.role === 'assistant'
-            ? 'border-b border-black/10 bg-gray-50 text-gray-800 dark:border-gray-900/50 dark:bg-[#444654] dark:text-gray-100'
-            : 'border-b border-black/10 bg-white text-gray-800 dark:border-gray-900/50 dark:bg-[#343541] dark:text-gray-100'
-        }`}
+        className={`group px-4 ${message.role === 'assistant'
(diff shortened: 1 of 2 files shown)


saliksik:
if the HEX code is formatted this way, the color highlight does not work

Screenshot (306)

Screenshot (307)

SyedMuzamilM:
Will check and fix.

SyedMuzamilM:
I think this is a code block should this also have a highlight after it

components/Chat/ChatMessage.tsx
```diff
@@ -212,6 +209,28 @@ export const ChatMessage: FC<Props> = memo(
                         </td>
                       );
                     },
+                    li({ children }) {
+                      const colorMatch =
```
SyedMuzamilM, line 213: This is how I can modify the code to check if it contains a code element and then check for the hex code. Is it ok to add here then we will get the color highlight if AI returns the hex color in code element or ``

``` typescript
const regex = /#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})/g;
  let colorMatch = regex
    .exec(
      String(children),
    );

  if (children?.length > 1) {
    Children.forEach(children, (child) => {
      if (isValidElement(child) && child.props.node.tagName === 'code') {
        console.log({ child });
        const match = regex.exec(child.props.children);
        if (match) {
          colorMatch = match;
        }
      }
    })
  }
```

mckaywrigley:
I think this is a really cool idea, but I don't think we will support this by default.

I would save this code for when plugins come on Chatbot UI. You can build a hex color preview plugin that users can add to their ui :)

SyedMuzamilM:
Awesome idea.

PR #292, opened 2023-03-29
feat: edit AI responses

RajSunner:
Hi, I've made changes that enable the AI response to be edited (using `@uiw/react-md-editor`). 

It needs refactoring in `index.tsx` and no doubt CSS tweaks will be needed to align the editor with the surroundings, but sharing the changes here early. Also, feel free to implement a different way if you're interested in the idea but not this execution!

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -41,7 +41,6 @@ interface Props {
   ) => void;
   onEditMessage: (message: Message, messageIndex: number) => void;
   stopConversationRef: MutableRefObject<boolean>;
-  lightMode: 'light' | 'dark'; 
 }
 
 export const Chat: FC<Props> = memo(
@@ -58,8 +57,7 @@ export const Chat: FC<Props> = memo(
     onSend,
     onUpdateConversation,
     onEditMessage,
-    stopConversationRef,
-    lightMode
+    stopConversationRef
   }) => {
     const { t } = useTranslation('chat');
     const [currentMessage, setCurrentMessage] = useState<Message>();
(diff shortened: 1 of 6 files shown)


srsholmes:
What is the use case for editing the api responses?

Jdban:
> What is the use case for editing the api responses?

OpenAI playground allows it. IMO it's useful when the AI gets something slightly wrong and you want to fix it, you can do it without having to add more messages and wait for the AI to correct itself.

Also if you're trying to get under max context size, you can trim older messages from the AI.

RajSunner:
> What is the use case for editing the api responses?

It works well with the copy functionality. If the response isn't quite right you can edit prior to copying. Also, you can make changes as needed to improve the rest of the conversation (by editing the context directly so the next response is even better / more accurate).

srsholmes:
Nice! 

Thanks for the explanation @RajSunner and @Jdban.

Much appreciated 👍

mckaywrigley:
Is there any particular reason why we're using a new library instead of just a basic text editor like editing user messages?

RajSunner:
> Is there any particular reason why we're using a new library instead of just a basic text editor like editing user messages?

Honestly, I didn't appreciate how easy it was to implement using the same approach to editing user messages! I've since removed the external lib and reverted back all changes in relation to that.

I've noticed that the current approach to editing is highly coupled with `handleSend`.  An improvement would be to allow messages to be edited without overwriting or removing all subsequent pre-existing messages. This links in with introducing basic CRUD functionality at a high level that covers #341 and #333, as well as this PR itself.

PR #293, opened 2023-03-29
use proxychains4 to access the OpenAI API in the command line.

zpeng2022:
## background
- I don't have a OPENAI_API_HOST
- I set the proxy global, but it always complains that it can't connect the OpenAI API
- I use `proxychains npm run dev` command, but it often gives errors.
- After using `proxychains4 npm run dev`, I got it to work.

- 
image

## proxychains4
```
proxychains4 npm run dev
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,21 @@ OPENAI_API_KEY=YOUR_KEY
 npm run dev
 ```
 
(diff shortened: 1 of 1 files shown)


mckaywrigley:
This looks like a great fix for your issue, but it's probably a bit too specific to merge in.

PR #295, opened 2023-03-29
Improve API error handling for better debugging

ryanhex53:
Previously, developers were only notified of a 'Bad Request' error when receiving a 400 response. 

image

However, with the new code, developers can receive detailed information about what happened during the API call. This allows for better debugging and error handling.

image


foxfog:
Thanks

jplopezmolinari:
Same here "Error"

PR #311, opened 2023-03-30
fix: small ui issue

SyedMuzamilM:
Removed the lines before the prompt and chat bars in the sidebar.

Added the light mode color for scroll down
## Before (Lines in sidebar without folder)
overlap
 button

## After (Removed the lines if there is no folder)
overlap-resolved


dasunNimantha:
@SyedMuzamilM After this update Scroll down arrow became invisible in the dark mode.

image

SyedMuzamilM:
Ok will fix this I think the color of the arrow is not visible.

PR #319, opened 2023-03-31
Feature: add link to openai account usage

rudolfolah:
* adds an external link that opens in a new tab to the OpenAI account usage page
* this is useful to see how many requests have been spent and the cost
* this becomes more useful when there is a choice of models

Screenshot 2023-03-30 at 9 33 25 PM


mckaywrigley:
On 2nd look I think the import statement for the icon is missing. Just fix that and I'll merge!

PR #328, opened 2023-03-31
feat: Add toast notification to surface OpenAI API errors to users

Jdban:
Adds a toast to let users know why their request failed
Should be a better user experience than silent failure

image


mckaywrigley:
I really like this! Eventually we may want our own custom toast component, but I think this is great for now. Well done.

PR #329, opened 2023-03-31
feat: Adds a loading spinner that replaces send icon while result is generated

Jdban:
Adds a loading spinner that replaces send icon while result is generated

Replaces icon when you hit send, goes away when a result is generated or an error happens

image
image


mckaywrigley:
fantastic add!

PR #330, opened 2023-03-31
feat: support import and export with prompts

superman66:
This PR adds the prompts field for data import and export.


SyedMuzamilM:
The code looks fine to me. But I think you are only handling the export part. You also need to implement the import part when there are prompts you need to handle import for that as well.

superman66:
> The code looks fine to me. But I think you are only handling the export part. You also need to implement the import part when there are prompts you need to handle import for that as well.↳

I think i'm responsible for implementing a crucial aspect of the project, which comprises of two parts. 

Firstly, I have set prompts to be stored in localStorage
image

which can be accessed through this link: https://github.com/mckaywrigley/chatbot-ui/pull/330/files#diff-b76d22796952ea80058b30849304cf51fbd7f7ff6fd47597976660a2565a7333R112-R123. 

Secondly, I have ensured that the data stored in localStorage for prompts can be imported to the State of the application by the user. 

image

This can be found here: https://github.com/mckaywrigley/chatbot-ui/pull/330/files#diff-aa98fd0757d0e1741503c50cfafb7726939d19819638bbe8e030a27adfec34a3R304.

If any sections are absent, kindly inform me.

SyedMuzamilM:
Sorry. I missed guy have skipped this part. Well everything looks fine to me then. Great Job.

mckaywrigley:
Alright finally got around to testing this and it seems to be working great!

LanceHaverkamp:
Where is this feature supposed to be? I see no option for import/export of prompts at all, at least not on the public website. Is it an older version, or ???

superman66:
@LanceHaverkamp 
When you export  conversions, it will includes prompts

image

Issue #207, opened 2023-03-26
Feature Request: Support ChatGPT Plugins


jiayao:
As a starting point, you could equip ChatBot with a memory of past conversations by integrating https://github.com/openai/chatgpt-retrieval-plugin/tree/main/examples/memory

kaaboaye:
LangChain has already created working plugins implementation. It could be used to inspire ChatBot's implementation.

https://python.langchain.com/en/latest/modules/agents/tools/examples/chatgpt_plugins.html

PR #331, opened 2023-03-31
Plugins (DRAFT)

cfortuner:
Started working on plugins.

very rough so far, but hoping to find collaborators on this!

Until openai adds a plugin api, the way I think we can support plugins in a simple way by doing two requests:

ex
 1. user submits a message

 2. We first try to use the enabled plugins

To try use a plug-in:

We need to ask an llm what to do next, given the conversation and a list of plug-in descriptions.

(In my tests I've found using a non-chat model like davinci and using the reAct prompting style works for this)

We should loop until the llm returns no plugin.

Each time the llm returns a plugin, we execute it and update the ui with the state of the plugin call.

3. Once step 2 is complete, we can include the result of the plug-in calls in a normal call to /api/chat to get the response

---

Since the user is supplying their api key, we could handle the plug-in logic on the client side for simplicity for now.


kaaboaye:
Just a friendly note. This project is using npm, not pnpm. Please make sure to use npm install when setting up this project.

Great job btw

cometbus:
> Started working on plugins.
> 
> very rough so far, but hoping to find collaborators on this!
> 
> Until openai adds a plugin api, the way I think we can support plugins is by doing two requests:
> 
> ex
> 
>     1. user sends message
> 
>     2. prompt 1 is used to determine what plugin to use
> 
>     3. if prompt 1 decides on using the plugin, we run the plugin
> 
>     4. prompt 2 has chat history and takes in user message and prompt 1 reply and then prompt 2 generates a response
> 
> 
> Open to other ideas!

I was trying to find a way to add customer helper files I wrote for a chatbot flask app that allows the bot to scrape websites and pdf files for text. I was trying to add commands I could execute from chat or buttons that would run the new features but I don't know enough coding and I just break the ui lol

bilalnawaz072:
Can you share pics how you used plugin?

mckaywrigley:
Very interesting! I’ll dig into this over the weekend.

cfortuner:
Updated the description with a slightly different approach.

Will give it a go tomorrow

kaaboaye:
Ref https://github.com/mckaywrigley/chatbot-ui/issues/207

tobi:
The plugin api is actually pretty set, and you can use it fine from the normal chatgpt api. 

The trick is that you you need to understand the OpenAPI spec and embed it into the system prompt. The LLM is fine tuned to be very sensitive to that and start invoking the plugins that it knows about. You can grab the [openapi endpoints](https://server.shop.app/openai/v1/api.json) linked from the ai-plugin.json of any conforming plugin. 

(i'd paste an example right now of the plugins prompt, but i'm in Italy and... well) 

the assistant will start queries to the plugin via `message: "to=plugin-name.openapi-endpoint-name\n{json}"`. I think 'name' may be undocumented. The response needs to be inserted as a chat message from `{"role": 'system', "name": plugin-name, "to=assistant\n{json}"`.  


Edit: Most of it is documented here: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb, except not the to= syntaxes which have to be a header to the message param. I assume these will eventually go into a nicer to parameter in future versions of the openai libraries. 


One request: once we get this, please allow a `DEFAULT_PLUGINS=` env variable for internally deployed chatbot-ui installations with automatic access to retrieval systems. Asking for a friend ;-)

bilalnawaz072:
Can you make a little bit video how you used and setup Prompt for plugin?

jpzhangvincent:
Is this ready to get merged? Interested in testing it out.

Issue #243, opened 2023-03-27
Deploy against Azure OpenAI Studio fails for ChatGPT API incompabilities


marcellodesales:
# 🤖 Service

* We have an Azure subscription with ChatGPT-3 and ChatGPT-4...
  * Cognitive Services -> Azure OpenAI Studio 
* We need to set the API URL to a different one

# 👽 API Differences

* HTTP Header for API KEY
* Message structure

# 🐛 Bugs

* We can't provide a different HOST, through Env Vars
* The API is hard-coded and, as a result, it can't be changed to a different set of URLs

> **NOTE**: The Azure OpenAI Studio completely broke the API contract defined by the official openai and so multiple changes are needed for `chat` and `model` APIs.
> * `Models API`: https://learn.microsoft.com/en-us/rest/api/cognitiveservices/azureopenaistable/deployments/list?tabs=HTTP#code-try-0
> * `Chat API`: couldn't find a reference, but it belongs to a given deployment as well

marcellodesales:
Found

https://github.com/mckaywrigley/chatbot-ui/blob/main/utils/server/index.ts#L8-L19

marcellodesales:
# 📸  Solution for Azure OpenAI API - Screenshots

<img width="1725" alt="Screenshot 2023-03-29 at 2 36 28 AM" src="https://user-images.githubusercontent.com/131457/228497954-796364e4-943f-4cd7-b046-4ec5d7b9da83.png">
<img width="1728" alt="Screenshot 2023-03-29 at 2 42 32 AM" src="https://user-images.githubusercontent.com/131457/228497965-b2c2a207-bc72-4202-95c7-efec1a064df7.png">

# 🛠️  Got it to work with Azure's OpenAI API

* Env var is not being properly passed from the docker-compose to Dockerfile to NextJS env vars.
* Azure's OpenAI API breaks the contract from the openai APIs
* The interface to send and receive messages is broken and needs to be changed
* HTTP Header for the API KEY is completely different from the openai version's

```console
$ git status
On branch feature/#243/use-azure-openai-studio
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   Dockerfile
        modified:   components/Chat/Chat.tsx
        modified:   components/Chat/ChatLoader.tsx
        modified:   components/Chat/ChatMessage.tsx
        modified:   docker-compose.yml
        modified:   next.config.js
[...]

PR #337, opened 2023-03-31
doc: update README.md for using Azure OpenAI Service

liby:
I made some local changes to support Azure OpenAI Service and found that a significant amount of code needs modification. This would greatly increase the code's complexity, and there are still some difficult-to-solve problems.

First, let's discuss the changes necessary to support Azure OpenAI Service:

1. Different API route for `/chat/completion`: `/openai/deployments/{deploymentID}/chat/completions?api-version=2023-03-15-preview`
2. Azure OpenAI Service does not have `/v1/models` API, so it needs to be replaced with the `/openai/deployments?api-version=2023-03-15-preview` API
3. There is an Auth Key mismatch. `headers` must include the `api-key`
4. [Some model names are different.](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/how-to/chatgpt?pivots=programming-language-chat-completions#model-versioning)

The existing problems are:

1. The current code cannot stop generation, which requires modifications for support. This might be due to the differences between OpenAI and Azure OpenAI results, with [but eventsource-parser did not make any special treatment for Azure OpenAI.](https://github.com/rexxars/eventsource-parser/blob/main/src/parse.ts#L122)
      - Azure OpenAI API output is missing a `\n` compared to OpenAI API output:
          image
          image
    a. [To get the /chat/completions to stop generating, I added the following code to utils/server/index.ts](https://github.com/mckaywrigley/chatbot-ui/issues/243#issuecomment-1491533867)
    b. [After switching to Azure, the conversation cannot be automatically stopped](https://github.com/Chanzhaoyu/chatgpt-web/issues/831)
2. Streaming output has issues; it's not possible to output word by word for an unknown reason.

In conclusion, I recommend using Azure OpenAI Service proxy to convert official OpenAI API requests into Azure OpenAI API requests. This approach is more elegant than modifying the code repository, as it avoids extensive code changes and doesn't require adding additional environment variables (which would be inconvenient for users).

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,10 @@ When deploying the application, the following environment variables can be set:
 If you do not provide an OpenAI API key with `OPENAI_API_KEY`, users will have to provide their own key.
 If you don't have an OpenAI API key, you can get one [here](https://platform.openai.com/account/api-keys).
 
+### Microsoft Azure Endpoints
+
+Please consider using Azure OpenAI service proxy to convert OpenAI official API request to Azure OpenAI API request. For example, [stulzq/azure-openai-proxy](https://github.com/stulzq/azure-openai-proxy) or [diemus/azure-openai-proxy](https://github.com/diemus/azure-openai-proxy).
+
 ## Contact
 
 If you have any questions, feel free to reach out to me on [Twitter](https://twitter.com/mckaywrigley).


mckaywrigley:
For simplicity in the official repo I’m going to keep it as-is. I think your case here is interesting, but probably a bit too specific to incorporate it here.

PR #356, opened 2023-04-02
remove unused chatbar/promptbar toggle code

bcullman:
This code is never called, (perhaps it is an earlier remnant of how to chatbar & promptbar were originally toggled)

(To be clear - the chatbar and promptbar still remain toggleable through buttons that are in place in index.tsx, here https://github.com/mckaywrigley/chatbot-ui/blob/main/pages/index.tsx#L675-L687 and 
https://github.com/mckaywrigley/chatbot-ui/blob/main/pages/index.tsx#L726-L738)


mckaywrigley:
nailed it. old remnant

PR #339, opened 2023-04-01
new feature: tree-like conversation

littlemoonstones:
I have added a new feature: a tree-like conversation, but my code is a little bit messy and there is a bug that causes the loading animation to show immediately. (Fixed it in https://github.com/mckaywrigley/chatbot-ui/commit/fc3f3cf6ef2b92aaee8bdf61bb2d5c8c2b05e675)

https://user-images.githubusercontent.com/32943414/229262252-3533d692-9eb6-4081-994b-636d14ef69ac.mov

PR #377, opened 2023-04-02
Import ChatGPT Conversations / Don't Overwrite Data on Import

RossHartmann:
This PR is meant to accomplish a few things:
- When importing data, merge the imported data and the existing data so that current data is not overridden.
- Add an additional import data format to cover ChatGPT conversation data (from chat.openai.com/chat).
- Add a script to export/save ChatGPT conversation data (using something like TamperMonkey, for instance).

I've added tests for cleaning and converting the ChatGPT data as well as for merging existing data into the imported data.


RossHartmann:
A couple of issues I noticed while importing some of my own ChatGPT conversations:
- ChatbotUI currently marks messages as "AI" and "You". It seems like these labels are just being applied to alternating message rather than respecting the message's "role". This causes problems when a ChatGPT conversation has multiple messages in a row from the "AI" (such as when using the browser plugin).
- Should we expand the metadata stored in a Message object in order to capture more of the nuance that is stored in ChatGPT data? (such as the browser plugin's citations, etc.)

bcullman:
@RossHartmann: There might be some overlap here with #339 - you might want to sync up with @littlemoonstones

RossHartmann:
> @RossHartmann: There might be some overlap here with #339 - you might want to sync up with @littlemoonstones

Good catch. Looks like the only overlap is in `chat.ts` where we both added "system" to the available options on the `Role` type.

I suppose whichever branch gets merged in second can resolve the conflict?

Beyond that it doesn't look like there are any conflicts (that I can see anyway).

RossHartmann:
> @RossHartmann: There might be some overlap here with #339 - you might want to sync up with @littlemoonstones

Good catch. Looks like there's an overlap in `chat.ts` where we both added "system" to the available options on the `Role` type.

Additionally #339 changes the way Messages are modeled in the Conversation interface.

I'm happy to either wait until #339 is merged and then adjust this branch or I can modify this branch to reflect the new Conversation/Messages model and #339 can adjust once this PR is merged.

(Either way, looks like we'll need to add a new export format based on the changes in #339 and then handle the necessary upgrading of the data format).

@littlemoonstones do you have a preference?

littlemoonstones:
> > @RossHartmann: There might be some overlap here with #339 - you might want to sync up with @littlemoonstones
> 
> Good catch. Looks like there's an overlap in `chat.ts` where we both added "system" to the available options on the `Role` type.
> 
> Additionally #339 changes the way Messages are modeled in the Conversation interface.
> 
> I'm happy to either wait until #339 is merged and then adjust this branch or I can modify this branch to reflect the new Conversation/Messages model and #339 can adjust once this PR is merged.
> 
> (Either way, looks like we'll need to add a new export format based on the changes in #339 and then handle the necessary upgrading of the data format).
> 
> @littlemoonstones do you have a preference?

I prefer #377 to be merged first because I believe #339 will take some time to be merged.

RossHartmann:
This PR is now pending #339 due to a fairly large Conversation model overhaul.

pages/index.tsx
```diff
@@ -61,51 +64,64 @@ const Home: React.FC<HomeProps> = ({
...
-        updatedConversation = {
-          ...selectedConversation,
-          messages: [...updatedMessages, message],
-        };
-      } else {
-        updatedConversation = {
-          ...selectedConversation,
-          messages: [...selectedConversation.messages, message],
-        };
+      let sendMessages: Message[];
```
littlemoonstones: The type of `Message` is causing the error when sending the request.
```
[OpenAIError: Additional properties are not allowed ('create_time', 'id' were unexpected) - 'messages.1'] {
  name: 'OpenAIError',
  type: 'invalid_request_error',
  param: null,
  code: null
}
```

littlemoonstones: I add another type to fixed it.
https://github.com/littlemoonstones/chatbot-ui/blob/8487bfa2e6c58d7388e62dfe07ecb704d746c8d0/types/chat.ts#L4-L14

RossHartmann: Thanks @littlemoonstones! I'll merge that in today.

utils/app/importExport.ts
```diff
@@ -24,41 +43,48 @@ export function isExportFormatV4(obj: any): obj is ExportFormatV4 {
...
   }
 
-
-  throw new Error('Unsupported data format');
+  if (!originalDataFormat) {
+    throw new Error('Unsupported data format');
+  } else {
+    return data;
+  }
 }
```
littlemoonstones: epxorting v5 data is susscessful, but import v5 data will raise Error.
image

littlemoonstones: I tried adding those code between line 82 and 83, and it's work:
```typescript
  if (isExportFormatV5(data)) {
    originalDataFormat = '5';
  }
```

RossHartmann: Okay cool I'll check this out, thank you!

pages/index.tsx
RossHartmann: Fixed

PR #380, opened 2023-04-02
:sparkles: add a script to automatically suggest translations

thomasleveil:
## TLDR

- Run `npm run update-translations`
- review changes
- commit what makes sense

## Motivation

Maintaining locale files is difficult because no dev is fluent in all supported languages. This PR adds a script that uses ChatGPT to provide suggestions for missing translations. It's then up to the developer to choose what translation suggestion is safe to add and commit.
When adding new UI text elements, this would provide at least an AI generated translation that may not be perfect. Humans could then submit PR to improve those new translations.


srsholmes:
silicon-valley

What an awesome use of AI. ❤️ Love this.  

Fantastic work @thomasleveil.

scripts/update-translations.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const response = await openai.createChatCompletion({
+    model: 'gpt-3.5-turbo',
+    temperature: 0,
+    messages: [
+      {
+        role: 'system',
+        content: `You are a translator and know all languages. 
+You are tasked with maintaining an application locale translation files expressed in JSON.
+The JSON content is composed of keys representing English label that must not change, and values that represent the tranlation of those labels.
```
ivanfioravanti: tranlation instead of translation

thomasleveil: Good catch ! fixed

thomasleveil:
rebased

thomasleveil:
rebased

PR #283, opened 2023-03-29
Adds a prettier formatting check to the github actions

Jdban:
This adds a formatting check via prettier to the github actions.

I noticed some PRs were a bit hairy with formatting changes, this would help alleviate that.

PR #395, opened 2023-04-03
chore: add husky and lint-staged for code quality checks

Jungley8:
Add Husky and lint-staged: 
1.  auto-format with Prettier and lint before commit (pre-commit hook)
2. run build before push (pre-push hook)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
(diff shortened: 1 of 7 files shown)


TommyE123:
I'm really sorry to say but I'm not really a fan of this approach as per the comments in #283.

I agree with @srsholmes and the short video he posted sums up why. [Precommit Hooks Are Bad!](https://www.youtube.com/watch?v=RAelLqnnOp0)

My personal opinion is that we should be looking to protect the main branch from PRs that could introduce poor code quality and style issues.

Issue #388, opened 2023-04-03
DEFAULT_SYSTEM_PROMPT not working


lzm0:
I'm using Vercel, and I configured a different system prompt as an environment variable in my project dashboard. However, the system prompt shown in a new chat remains the original.

I would like to know if `DEFAULT_SYSTEM_PROMPT` is working for others.

chepidev:
hi, just wanna know, why is the script wont generate any respon? its still working yesterday

jponline77:
Doesn't work for me.  I've been working with environmental variables all morning and it seems something is broken with the utils/app/const.ts, It seems that the the environmental variables get erased after the client starts.  I was working on creating an environmental variable to change the title and it just won't work.  However, I was also using system prompt and it has the same problem.

jponline77:
I root caused the issue as an issue with environmental variables not being visible at runtime in Next.  I created a pull request to fix the issue.

PR #399, opened 2023-04-03
Fixed issue with SYSTEM_PROMPT environmental variable not working.  Created ability to change the title of the Chatbot UI with an environmental variable.

jponline77:
I found that the system prompt was not working.  It seems that it needs to be NEXT_PUBLIC for it to be visible at runtime.  Also, I added the ability to change the title for people who are running internal chatgpt sessions and want a different name.

### DIFF
--- a/pages/api/google.ts
+++ b/pages/api/google.ts
@@ -122,7 +122,7 @@ const handler = async (req: NextApiRequest, res: NextApiResponse<any>) => {
         messages: [
           {
             role: 'system',
-            content: `Use the sources to provide an accurate response. Respond in markdown format. Cite the sources you used as [1](link), etc, as you use them. Maximum 4 sentenes.`,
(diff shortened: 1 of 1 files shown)


lzm0:
This works for me.

jponline77:
> This works for me.

What environment are you using it in?  I am using it as a standalone running directly with npm command.  I suspect that it may work in other environments like docker.  It has to do with the fact that Next does not allow system variables to be present at runtime.  When I add NEXT_PUBLIC_ in front of the environmental variable it starts working.  See https://nextjs.org/docs/basic-features/environment-variables.  It seems other users have the same problem see #388

mthurmond:
Thanks @jponline77, replacing DEFAULT_SYSTEM_PROMPT with NEXT_PUBLIC_DEFAULT_SYSTEM_PROMPT in both the .env.local and the const.ts files fixed this for me. 

Note that this is referring to my local testing environment, where I'm using the next.js script 'npm run dev' to test. I'm not using docker or anything. I believe this same solution works when using Vercel to deploy the app too.

PR #407, opened 2023-04-04
fix: auto popup prompt modal when creating

MatriQ:
QA screen recording:

https://user-images.githubusercontent.com/1320925/229700967-aeddb0a9-c745-4996-9c66-5abf51004933.mp4


pages/index.tsx
```diff
@@ -496,8 +497,6 @@ const Home: React.FC<HomeProps> = ({
   // PROMPT OPERATIONS --------------------------------------------
 
   const handleCreatePrompt = () => {
-    const lastPrompt = prompts[prompts.length - 1];
```
MatriQ, line 499: Remove it because it's not used in the scope.

PR #414, opened 2023-04-04
google search beta

mckaywrigley:


mckaywrigley:
Not perfect but I want to get this out so people can begin working on it and providing feedback.

rossman22590:
how does this work?

MarioGiancini:
@mckaywrigley @rossman22590 

~~It looks like this is enable via `const [showPluginSelect, setShowPluginSelect] = useState(false);` and is hardcoded to `false`, so the Plugin dropdown doesn't show.~~

~~I have a fix for this here: https://github.com/mckaywrigley/chatbot-ui/pull/477~~

Sorry, ignore the above. Complete missed the `IconBolt` button, See here: https://github.com/mckaywrigley/chatbot-ui/pull/477#issuecomment-1500929792

PR #427, opened 2023-04-04
Fix 416 copy button on code block

### LINKED ISSUE #416: "Copy code" button not appearing on codeblocks
I noticed that the "Copy code" button for generated codeblocks is not appearing. I've tried multiple instances of asking the AI to generate code and it won't appear no matter what I try, this is both in a local build and in live production (Vercel) as of the latest commit.

image

thomasleveil:
As reported on #416, the _Copy code_ button does not always shows up on some code blocks, specifically when the content language is not recognized.

## Before fix

image

## After fix

image

Fixes #416


Jdban:
Thank you!

PR #442, opened 2023-04-05
fix: Modal not scrolling on small devices

### LINKED ISSUE #353: Save button invisible on mobile prompt creation
Hiya folks,

Prompt creation isn't possible on mobile devices because the save button is missing:
mobileprompt

SyedMuzamilM:
Prompt Modal was not scrolling on small devices and the button was hidden. And Some elements were on top of the modal.

## Before
overlap

## After
overlap-resolved

Close #353


dornfeld-capital:
Nice work Syed. Why not make the modal full width and full height?

mckaywrigley:

SyedMuzamilM:
Yeah sure.

Issue #267, opened 2023-03-28
Feature Request: add options for temperature


TiancongLx:

alexzadeh:
+1

ivanfioravanti:
+1

ivanfioravanti:
I’m working on adding this one at the beginning of a conversation, like for System Prompt. It will be in my fork soon, beta, I started from 0 on next.js & c.

PR #446, opened 2023-04-05
Temperature added as parameter

ivanfioravanti:
Temperature added as initial conversation parameter as per issue #267 

Never used next.js before so please review this one 😂


mckaywrigley:

ivanfioravanti:
Here it is: #513

PR #458, opened 2023-04-06
feat: add font selector

dornfeld-capital:
Screenshot 2023-04-06 at 2 51 31 PM

Add font selector for
- Inter
- Lexend
- Open Sans
- Roboto
- Ubuntu


mckaywrigley:

Issue #359, opened 2023-04-02
Can I use Azure OpenAI Service


TiancongLx:
Using it from unsupported region like CN and HK

casaltenburg:
I think it’s a separate feature , because the API for azure openAI service is a little different.

vladiliescu:
@TiancongLx See [my fork](https://github.com/vladiliescu/azure-chatbot-ui) for this, I'm using it locally but Docker is most likely broken. I'm planning to keep it in sync with the main repo, still figuring out the best way to do this.

TiancongLx:
> I think it’s a separate feature , because the API for azure openAI service is a little different.

An iOS GPT client called OpenCat can customize the API URL, which I think is a wonderful feature since the openai api key will be banned frequently and the account will be suspended with no reason. It is so annoying to using AI service in unsupported regions.

leodambrosi:
@vladiliescu thank you. Does your version still require openai key? I tried to make it run but without openAI won't let me chat. When added, I get a resource error. Thank you

vladiliescu:
@leodambrosi Of course, you will still need an Azure OpenAI key.

leodambrosi:
@vladiliescu thanks. Can you help me understand why "of course"? Azure open AI works independently from OpenAI API. If you have the OpenAI API, why should you use azure api?

vladiliescu:
@leodambrosi Depending on which service you choose to use (OpenAI or Azure OpenAI) you'll need an API key from that respective service. If you're using Azure OpenAI you'll also need a few more things, see [here](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/chatgpt-quickstart?tabs=bash&pivots=rest-api#retrieve-key-and-endpoint) for some details on that.

leodambrosi:
@vladiliescu I have both Azure API an OpenAI API. My question is: why is the OpenAI API still requested on your Azure API based fork?

vladiliescu:
@leodambrosi Cool, take a look at [env.local.example](https://github.com/vladiliescu/azure-chatbot-ui/blob/dev/.env.local.example) to see what you need to configure, the values are referenced in the [readme](https://github.com/vladiliescu/azure-chatbot-ui/tree/dev#configuration) too. You don't need an OpenAI API key, have you filled in `env.local` as per the docs and it still doesn't work for you?

leodambrosi:
[...]

PR #460, opened 2023-04-06
Add support for Azure OpenAI

### LINKED ISSUE #243: Deploy against Azure OpenAI Studio fails for ChatGPT API incompabilities

itbm:
Closes #243 #359 #479

### DIFF
--- a/pages/api/models.ts
+++ b/pages/api/models.ts
@@ -49,8 +49,9 @@ const handler = async (req: Request): Promise<Response> => {
 
     const models: OpenAIModel[] = json.data
       .map((model: any) => {
+        const model_name = OPENAI_API_TYPE === 'azure'? model.model: model.id
         for (const [key, value] of Object.entries(OpenAIModelID)) {
-          if (value === model.id) {
+          if (value === model_name) {
             return {
               id: model.id,
               name: OpenAIModels[value].name,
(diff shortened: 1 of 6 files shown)


ivanfioravanti:
Great one!

types/openai.ts
```diff
@@ -7,7 +9,9 @@ export interface OpenAIModel {
 
 export enum OpenAIModelID {
   GPT_3_5 = 'gpt-3.5-turbo',
+  GPT_3_5_AZ = 'gpt-35-turbo',
```
brunodoamaral, line 12: Hi. First, thanks for providing this reference implementation!

But note that Azure provides a way to name the model, what they call "deployment name" ([step 6 of this link)](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/how-to/create-resource?pivots=web-portal#deploy-a-model):

> The model name will be used as the deployment name via OpenAI client libraries and API.

Therefore, we need a way to specify custom model names instead of hard-coding them.

pages/api/models.ts
lewcpe, line 1: The `model.id` returned from the `/openai/deployments` is user defined not system config. The line 53 should be `if (value === model.model) {`

e10101:
Waiting for this too. :)

mckaywrigley:

PR #477, opened 2023-04-08
Auto-show Plugins Dropdown If Google API Keys Are Set

### LINKED ISSUE #476: Show Plugin In Dropdown Only If API Keys Are Given
### Overview

Currently, the first plugin was for Google search and is in beta via #414. 

We should update the components to only show the plugins in the dropdown if we've set our `.env` keys for them, in this case the Google API plugin.

MarioGiancini:
### Description
Passes `serverSidePluginKeysSet` prop to all components that would need it to automatically check if we should show a plugin dropdown if Google API keys are set. This fixes the hardcoded `showPluginSelect` state.

Closes #476

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -29,6 +29,7 @@ interface Props {
   models: OpenAIModel[];
   apiKey: string;
   serverSideApiKeyIsSet: boolean;
+  serverSidePluginKeysSet: boolean;
   defaultModelId: OpenAIModelID;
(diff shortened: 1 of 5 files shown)


MarioGiancini:
Realizing now that this may not be needed, since the `IconBolt` is setting this 🤦‍♂️ 

Screenshot 2023-04-08 at 9 38 50 AM

MarioGiancini:
Not needed

PR #481, opened 2023-04-09
feature(devcontainer): api service with local llama models

mkellerman:
This is to add an api service to provide an api to llama.cpp using fastapi, and replicate openai endpoints for local testing.

This is an additional to the devcontainer pr #480


shep920:
pls

PR #484, opened 2023-04-10
feat: delete single message

L-Ryland:


mckaywrigley:

it *should* be much easier to develop on now.

Issue #218, opened 2023-03-27
firefox cannot type chinese in the message bar


MoXuanSun:

d3vw:
Me too

ryanhex53:
The same goes for me, whether it's the desktop version of Firefox or on my phone, I can't input Chinese characters.

ryanhex53:
I'm trying to fix this.

ryanhex53:
PR #505 Fixed this issue

ryanhex53:
@MoXuanSun The issue has been resolved in [my fork](https://github.com/ryanhex53/chatbot-ui), but the pull request has not yet been merged into this repository.

PR #253, opened 2023-03-28
fix: resolve Enter event conflict during CJK IME

liby:
## Before
https://user-images.githubusercontent.com/38807139/228122503-e25f2666-c43f-4521-9cb4-bd1dd4733ea1.mp4

## After
https://user-images.githubusercontent.com/38807139/228121948-6bd8ccd4-d554-456b-9576-4aeef2df7641.mp4

PR #505, opened 2023-04-11
[bugfix] Firefox experiences CJK IME issues due to CompositionEvent.

ryanhex53:
Related to PR #253, fixes an issue with Chinese input not working properly in Firefox. Fixes issue #218


ryanhex53:
Before
before

After
after

liby:
1. The issue was introduced by #4, not #253. This can be confirmed by looking at the [issue](https://github.com/mckaywrigley/chatbot-ui/issues/218) creation time and the PR merging time.
2. The PR doesn't explain why it causes the inability to input Chinese characters:
    - The main reason for this issue is the differing behavior of the `onCompositionEnd` event in Firefox compared to other browsers. The specific reason can be found at https://github.com/w3c/uievents/issues/202.
    - TL;DR, in Firefox, the "input" event is fired after "compositionend".
3. There are two potential issues when using the `onKeyUp` event to handle Chinese input:
   1. Delayed response: The `onKeyUp` event is triggered when a key is released, which may lead to delays in determining the input state. For example, when selecting a suggested word shown by the input method, there may be a delay during the selection process after pressing the space key.
   2. Inaccurate input state: In practice, to accurately handle the input method editing state, it's important to know whether the current input is input method editing or other input. In some cases (such as continuous keypresses), the `onKeyUp` event may be triggered before input event handling finishes, leading to inaccuracies in determining the input method editing state.

ryanhex53:
@liby Fine, Could you fix your code to solve the problem of inputting Chinese characters in Firefox? Because as long as `onCompositionStart` or `onCompositionEnd` is used, inputting cannot be done normally in Firefox. It took me several hours to figure out the problem and get it to a usable and functional state.

ryanhex53:
closed by mistake, should reopen this.

PR #507, opened 2023-04-11
Vary response tokens depending on request size

itbm:
This calculates `max_tokens` based on the number of request tokens (with a minimum of 25% of token limit). This enables larger responses from `gpt-4` and `gpt-4-32k`.


itbm:
Merge conflicts resolved

Issue #498, opened 2023-04-11
Error message "The API deployment for this resource does not exist" when using Azure OpenAI Service


jony4:
```
The API deployment for this resource does not exist. If you created the deployment within the last 5 minutes, please wait a moment and try again.
```

jony4:
Actually, I have deployed the OpenAI model and service already.

jony4:
Perhaps this parameter was omitted.

reference -> https://github.com/sashabaranov/go-openai/blob/master/config.go#L60

gchust:
I was facing the same error, I fixed it by setting the following environment var for azure

```diff
DEFAULT_MODEL | gpt-3.5-turbo | The default model to use on new conversations, for Azure use gpt-35-turbo

---gpt-3.5-trubo
+++gpt-35-turbo

```

Azure should use **gpt-35-turbo**

jony4:
no, this is deployment id is from here: https://learn.microsoft.com/zh-cn/azure/cognitive-services/openai/reference#completions which input by yourself

<img width="1328" alt="image" src="https://user-images.githubusercontent.com/13718575/231163751-d129fa8a-ca01-462c-9d63-e076fceb189b.png">

I just test with "gpt35", and no more error message. lol

jony4:
pr https://github.com/mckaywrigley/chatbot-ui/pull/499

gchust:
> no, this is deployment id is from here: https://learn.microsoft.com/zh-cn/azure/cognitive-services/openai/reference#completions which input by yourself
> 
> <img alt="image" width="1328" src="https://user-images.githubusercontent.com/13718575/231163751-d129fa8a-ca01-462c-9d63-e076fceb189b.png">
> 
> I just test with "gpt35", and no more error message. lol

You're right, I have renamed the deployment name same as the model name lol...

itbm:
@jony4 @mckaywrigley PR #499 isn't required and adds complexity. The existing solution automatically translates `gpt-35-turbo` / `gpt-4` / `gpt-4-32k` to your deployment name. See https://github.com/mckaywrigley/chatbot-ui/blob/main/pages/api/models.ts#L53.

If you opened the web UI before setting `DEFAULT_MODEL=gpt-35-turbo` it may have cached `gpt-3.5-turbo` in your conversations. Exporting your data, fixing the JSON file, clearing your cache and re-importing data will probably solve it _(or just clear your cache if you don't want to keep conversations)_. The exported JSON should have:

[...]

PR #508, opened 2023-04-11
Fix model switcher for Azure

itbm:
Reverts mckaywrigley/chatbot-ui#499

I suggest reverting this as it isn't necessary and breaks model switching between GPT-3.5, GPT-4 and GPT-4-32K in the UI. Should be merged with #509.

See https://github.com/mckaywrigley/chatbot-ui/issues/498#issuecomment-1503808802


kleptog:
This isn't correct. The deployment name in the URL is unrelated to the model name and must be configured. I don't think there is a way to get a list of available deployments unless you can access the Azure API.

itbm:
@kleptog it is correct. The azure api returns both the model name and the deployment name. The code already checks against the model (model field), then translates it to the deployment name (id field) for use in the URL.

See https://github.com/mckaywrigley/chatbot-ui/blob/main/pages/api/models.ts#L53.

With this PR it doesn't matter what the deployment is called, as it looks for the model and updates the URL accordingly.

itbm:
@mckaywrigley I’ve fixed merge conflicts. Would it be possible to get this reviewed and merged?

ivanfioravanti:
This works! Great job @itbm

brandco:
It looks like the Azure deployment name would need to match the model name for this to work. Is that correct? 

I'm not sure why someone would use a deployment name that doesn't match the model name in setting up Azure OpenAI, but it is possible. Perhaps this is useful for fine tuned models, I'm not sure.

itbm:
With this PR they can be different. Chatbot UI will use the model name internally up until it makes the request to Azure, at which point it will use the deployment name in the URL.

tobigue:
Currently always the default model (configured in `AZURE_DEPLOYMENT_ID`) is used instead of the model the user picked in the interface when using Azure as backend, so would be great to get this merged! :)

Kush-agra-wal:
> It looks like the Azure deployment name would need to match the model name for this to work. Is that correct?
> 
> I'm not sure why someone would use a deployment name that doesn't match the model name in setting up Azure OpenAI, but it is possible. Perhaps this is useful for fine tuned models, I'm not sure.

Still facing this issue, when deployment name is not = model name the default model assigned at start of conversation is not assigning the model id from azure instead gives fallback one. if one of the deployments is named gpt-35-turbo it is able to work fine for other deployments which dont have name same as model name

PR #509, opened 2023-04-11
Set default model correctly for Azure

itbm:
This sets the correct default model for Azure to prevent issues like #498 if `DEFAULT_MODEL` isn't set when using Azure OpenAI.

Should be merged with #508.


itbm:
Merge conflicts resolved

itbm:
@mckaywrigley would it be possible to get these 2 PR's merged so we can get Azure working?

ivanfioravanti:
This works! Great job @itbm

PR #557, opened 2023-04-14
fix: google plugin token calculation

dotneet:
Because the tokens was not calculated, most searches with multibyte characters resulted in errors.
With this fix, it now works correctly.

It was not possible to use Edge Runtime because some functionality depends on the node api.

If the limit is not appropriate, please change it.
 - 400 tokens per source
 - 2000 tokens limit for whole sources.


mckaywrigley:
just need to fix the prompt. the sourceTexts needs to be formatted as they were previously

PR #560, opened 2023-04-14
feat: next-auth -  email pattern matching only

dotneet:
This is very simple NextAuth implementation.
But for some people including me this will be enough.

This implementation uses only JWT and email patterns for authentication.


pdgxiaofei:
hello
Excuse me, how do you need to use this, I inherited your branch, and the startup project authentication does not work. I have filled in the required parameters.
Thank you very much if you can answer

kyleroberts-ai:
> hello
> Excuse me, how do you need to use this, I inherited your branch, and the startup project authentication does not work. I have filled in the required parameters.
> Thank you very much if you can answer

Have the same problem

dotneet:
Adding console.log to middleware.ts would provide insight.

dotneet:
@pdgxiaofei @QuickResolveAI 
sorry, I've fixed the problem that maybe you've encountered.
please try again.

Issue #566, opened 2023-04-15
Feature Request: Multi-line input


j4shu:
Wondering if a way to toggle explicitly single line + multi line would be possible, i.e. like slack. Ctrl/Cmd + enter to send message in multi-line mode

KhanKudo:
You can always use Shift + Enter to create a new line, without sending the message.
But a specific mode toggle like a button is something I personally have never seen, would be odd.
Changing it to default Enter as new-line and Ctrl+Enter as send could be confusing and unnecessary for most users.
If you have a suggestion on what this would look, I'd be happy to create a PR.

j4shu:
It is a pretty common practice, i.e. in Slack: https://slack.com/help/articles/115005523006-Set-your-Enter-key-preference and Microsoft Teams: https://camerondwyer.com/2018/12/22/do-you-hit-enter-in-microsoft-teams-for-a-new-line-and-accidentally-post-partial-message-try-this-hack/. It's usually a setting instead of default so it wouldn't affect most users.

KhanKudo:
I see what you mean. I'll take a closer look at it. I have a possible implementation in mind.

KhanKudo:
I believe this #581 is what you are looking for.
If it's not correct, I am happy to make any adjustments.

PR #419, opened 2023-04-04
feat: clean up sidebar

dornfeld-capital:
Screenshot 2023-04-04 at 12 27 48 PM

Screenshot 2023-04-04 at 12 27 55 PM

- Update add chat icon
- Add search icon to search input
- Improve conversation options layout
- Add toggle button for settings

PR #581, opened 2023-04-16
Added multi-line input option (also fixed theme setting not saved)

KhanKudo:
Implemented Feature Request #566.
By default, messages are sent using Enter.
In multi-line mode, messages are sent using Ctrl+Enter. Enter alone creates a newline.

In the process also added missing localStorage.setItem call for theme setting

### DIFF
--- a/components/Chat/ChatInput.tsx
+++ b/components/Chat/ChatInput.tsx
@@ -43,7 +43,7 @@ export const ChatInput = ({
   const { t } = useTranslation('chat');
 
   const {
-    state: { selectedConversation, messageIsStreaming, prompts },
+    state: { selectedConversation, messageIsStreaming, prompts, lineMode },
 
     dispatch: homeDispatch,
   } = useContext(HomeContext);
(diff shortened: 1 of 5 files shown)


mckaywrigley:
We use shift+enter for a new line which tends to be pretty common UX, so keeping that for now.

KhanKudo:
I do agree that the current method is the most common implementation.
However, given that this change keeps the current behaviour unchanged.
It simply allows users with different preferences to adjust the app to their liking.
Perhaps the placement of the toggle is too dominant, it should be behind a settings menu, which already seems to have a PR #419
Generally speaking however, the more control Users have, the better. At least that's my opinion.
I must say I completely understand your opinion. I think the PR is worth a second thought about it, but I perfectly understand if you don't feel it doesn't fit.

jshu-git:
This feature also helps immensely to send formatted code to the ChatCompletion endpoint.

Consider the following snippet:
```
# Program to generate a random number between 0 and 9

# importing the random module
import random

print(random.randint(0,9))
```

Without multi-line, copying and pasting this code as message and sending formats the message like so:

\`\`\`# Program to generate a random number between 0 and 9 # importing the random module import random print(random.randint(0,9))\`\`\`

This has shown to affect the quality of code generation for chat completion requests, as OpenAI models are trained on properly formatted markdown. With the current implementation, in order to send the code in the correct format, the user needs to paste the code snippet, then add manual line breaks with shift+enter after each line.

Having a toggle for multi-line input resolves this issue. It would not have to be default either, just available for users who primarily use the chat for code generation.

PR #582, opened 2023-04-16
fix: dont hide search bar on no results

Anerco:


mckaywrigley:
Just pushed a fix for this!

PR #585, opened 2023-04-17
feature: new class PromptScrappers introduced

mirkan1:
As shown below picture this PR implements scraper functionality that works on the fly and sets it into a prompts area

image


mckaywrigley:
Love this, but needs a way to set the custom download link.

Jdban:
I tested this out. I didn't like how it basically dumped 150 prompts into my prompt list and there was no easy way to delete them.

I'd suggest thinking about the use cases a bit more and seeing about implementing a friendlier UX.

For example, I could see someone wanting to add a couple prompts from here, but almost never want all of them.

mirkan1:
Both of you are right. I better implement a component for expected functionalities

mckaywrigley:
> Both of you are right. I better implement a component for expected functionalities

Have you looked into building your own custom tool that helps convert different prompt structures into our own import format? Could be a useful tool that users would want to use. I think @Jdban brings up good points that this might be too bespoke for native implementation here while still being a useful tool.

mirkan1:
I can finish writing the scrapper component that I mentioned but I figured there is an import data button that pretty much does the same thing.

PR #615, opened 2023-04-20
Add new text-to-speech feature for reading new messages

MalTarDesigns:
This code implements a new feature that enables text-to-speech reading of newly generated messages. The feature includes a toggle for the audio, which can be found in the top right of the user interface next to the clear all messages button. The audio can be enabled or disabled by clicking on the volume icon. The system will automatically read out new messages when the audio is enabled, and a new message is generated in the selected conversation.


PaulShroom:
https://github.com/C-Nedelcu/talk-to-chatgpt

Could this be useful for TTS implementation?

PaulShroom:
A play/stop button next to each text would be cool with voice speed and type options.

PR #617, opened 2023-04-20
Making Azure deployments work with model dropdown.

mathesz:
No need for AZURE_DEPLOYMENT_ID env since it's already populated in the model.


itbm:
A couple of other changes are required as part of this to get it fully working. It needs the fallback models to be set properly. 

I opened PRs to fix it a couple of weeks ago but they've not been merged, see #508 and #509.

PR #619, opened 2023-04-20
Add animated typing cursing

anthonypuppo:
Screenshot 2023-04-20 at 8 17 35 AM

Adds an animated typing cursor to bot replies while the message is loading/streaming. Works across the different output formats (codeblocks, tables, etc.).


mckaywrigley:
love this!

PR #627, opened 2023-04-22
styles : max width is  pretty

jiangxianfu:


pdgxiaofei:
image
image
Hello
Please look at the two screenshots above, it seems that the adaptation of the multi-button is still not perfect.

jiangxianfu:
> image image Hello Please look at the two screenshots above, it seems that the adaptation of the multi-button is still not perfect.

image
i can see you button, where ?

pdgxiaofei:
This copy and edit delete button appears when you click on your conversation with the mouse.

pdgxiaofei:
I mean, the delete button should be the same as before, it's best not to be too close to the right, otherwise the operation on the mobile will not be so perfect.

PR #645, opened 2023-04-25
Fix OpenAI model context length error by updating condition for maximum context length

ziizii:
It appears that the tokenizer in a third-party library is counting the tokens slightly differently than GPT, leading to a context length error:

```OpenAIError: This model's maximum context length is 4097 tokens. However, you requested 4170 tokens (3170 in the messages, 1000 in the completion). Please reduce the length of the messages or completion.```

To prevent this error, I updated the condition for checking the maximum context length to include a 10% buffer. This ensures that the maximum context length is not exceeded, while also accounting for any discrepancies in token counting by the tokenizer.

Note that a buffer of 0.95 would probably be enough to prevent the context length error in most cases. However, I set the buffer to 0.9 to make the code more robust and prevent any edge cases where the token counting may be even more different than expected.


dotneet:
This is the correct way to calculate tokens on chat completion API.
https://github.com/dqbd/tiktoken/issues/23#issuecomment-1483317174

ziizii:
I'm getting that error for a reason.

wallter:
@ziizii, your solution is commendable for eliminating the error/bug. However, @dotneet raises a valid point that accurately counting the number of tokens used is preferable to incorporating a 10% buffer. While the 10% buffer method does work, it results in a loss of 10-9% of the message's total input to the model.

Checkout how it can be implemented: https://github.com/dotneet/smart-chatbot-ui/blob/main/pages/api/chat.ts#L26

Issue #657, opened 2023-04-27
Show tokens consumed


multifolio:
I would like to see how many tokens have been consumed by each prompt and the total amount of tokens consumed be it per conversation or in total. This can be helpful to track costs. Since the the openAi response is streamed, it's hard (for me) to get that piece of information.

PR #664, opened 2023-04-28
Add token count and price estimation

dqbd:
Closes #272 #657 #128 

This uses @dqbd/tiktoken, which is a WASM port of openai/tiktoken, which should be more aligned with the actual token cost. The count logic is the same as in https://tiktokenizer.vercel.app/. I'm also only importing `cl100k_base`, which should match GPT-3.5 and GPT-4 models

For calculation itself, I'm using BigNumber.js to avoid floating-point arithmetic shenanigans and imprecision.  

https://user-images.githubusercontent.com/1443449/235177467-501f1a6a-0a0e-4ce9-a8f8-98c94aa36054.mp4


KarimMesallam:
I believe that the current implementation for token calculation might not accurately account for token usage, as it should consider the entire chat history for each subsequent user prompt. This is because the full chat history is sent as context for every new user prompt. In the current approach, the displayed token cost may be significantly lower than the actual consumption.

Furthermore, it's important to distinguish between token costs for prompts and completions in GPT-4, as they vary.
Prompt: $0.03 / 1K tokens | Completion: $0.06 / 1K tokens

dqbd:
Hello @KarimMesallam,

the current implementation does take into account the chat history, as seen here: https://github.com/mckaywrigley/chatbot-ui/pull/664/files#diff-df165e72d8963e70352e718d811658cbb15cdab68cd286708371a39867174347R40, as well as the system prompt. 

This PR does not really address tracking of the completion costs, only the prompting cost. General usage tracking of costs is definitely a good idea for a new PR though 😄

rdvo:
yeah can someone make it calcualte both?

espositodaniele:
How would you ideally implement client-side counting for tracking the cost of completion?

PR #661, opened 2023-04-28
feature: Multiple Database Support + Multiple Accounts Support + Data Syncing

jorge-menjivar:
## Relational Databases
Uses [TypeORM](https://github.com/typeorm/typeorm) to implement support for PostgreSQL and MySQL.
Support for MariaDB and CockroachDB is suspected but untested.

## CouchDB
Includes the work done in #561 which implements support for CouchDB.

## Multiple Account Support
Includes the work done in #560 to implement support for multiple user accounts in a single device **when used in conjunction with a relational database** as the storage method.

## Syncing Across Devices
By using a relational database as the storage method, now you can save and sync all data, excluding API Keys, across devices, as long as you use the same email to log in in each device.

A live version of this work, using PostgreSQL on google cloud as storage, can be found [here](https://chat.jmenjivar.com).

### Closes the following Issues
closes #630 
closes #620
closes #452
partial #659
partial #654

Plus a few closed by the work of #561 
closes #461
closes #182
closes #147
partial #367

PR #672, opened 2023-04-29
fix: client side token counting and undefined requestLimit

jorge-menjivar:
This pull request also includes #661 

- Fixes issue where the token count on the client side is not the same as the token count returned by the OpenAI API. [Reference](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb) (See section 6 for the approach used to fix this issue.)
- Fixes issue where the updated model passed to [handleUpdateConversation](https://github.com/mckaywrigley/chatbot-ui/blob/fa3f6e93bbe0d1ff9f208ddefae6fc7dfb738dc7/pages/api/home/home.tsx#L210) does not include `maxLength`, `tokenLimit`, and `requestLimit` when the conversation model is changed from the UI.


jorge-menjivar:
Below is an example json file that you can use in [chatbotui.com](https://www.chatbotui.com) vs the [deployment above](https://chatbot-41vuaeuc6-mckay-personal.vercel.app/) to test the error in the token count. Simply type "another one" after importing the json file.

Rename to test.json and import:
[test.txt](https://github.com/mckaywrigley/chatbot-ui/files/11469062/test.txt)

mmtmn (changes requested):
envs in pull request

Issue #383, opened 2023-04-02
Feature Request: Markdown export 


Dkeds4MX:
The option to export conversations to markdown format would be great

peterkaminski:
Another vote for markdown format export.

It should be built in, but for a model of how the UI and export format might look, check out [pionxzh/chatgpt-exporter: Export and Share your ChatGPT conversation history](https://github.com/pionxzh/chatgpt-exporter).  It's a Tampermonkey userscript that installs an export button on the OpenAI ChatGTP web interface.

peterkaminski:
I've submitted PR #684, which adds an "Export Markdown" button that exports conversations in Markdown in a zip file.

boozedog:
@mckaywrigley I just did a pull of this commit https://github.com/mckaywrigley/chatbot-ui/commit/369087057e0d90a38188ef06a6af0d3b685a81ee and started the app, not seeing the option to export to markdown? Any ideas? Thanks!

PR #684, opened 2023-05-01
feature: Markdown export

peterkaminski:
This PR adds Markdown export of conversations, addressing #383.

The current code works and allows you to export conversations as Markdown. It does not currently export prompts or metadata. A question: could it be merged as-is, or should we work on exporting prompts and metadata?

I changed the export filenames to be a little more informative, and to sort well as the user does exports over time. The format I used:

- chatbot_ui_export_20230501_data.json
- chatbot_ui_export_20230501_markdown.zip

Using this format, the "chatbot_ui_export_*" files cluster together, are sorted first by date, and then second by type of export.

Not yet addressed:

- export of prompts or metadata
- customization of export filenames (with simple templating, for instance)
- allow user to select a subset of all conversations for export (similar to the capability of [ChatGPT Exporter](https://github.com/pionxzh/chatgpt-exporter) for instance)
- display a spinner or progress bar or something while markdown zip file is assembled (probably not needed unless there are lots of conversations)

Also to note, the current code uses [adm-zip](https://github.com/cthackers/adm-zip), which seems fine, but we may want to look at different zip libraries such as [JSZip](https://stuk.github.io/jszip/). I briefly tried JSZip but didn't get the async call working, so I switched to adm-zip. A small downside of adm-zip is that it tries to include `fs`, which is used for e.g. Electron but not browsers, so a fallback needs to be specified in `next.config.js`.

Screenshot of the "Export Markdown" button in context as implemented:

Chatbot UI buttons at lower left, with Export Markdown button highlighted


peterkaminski:
I've added support for folders in the Markdown chat export in commit https://github.com/mckaywrigley/chatbot-ui/pull/684/commits/2209363145a94364440db4272bb2a8f7802266cb.

I've edited my first comment above to remove folders as a missing part of the PR.

PR #685, opened 2023-05-02
system-only mode

jmagly:
this should likely be a more robust feature. this restores GPT-4 fidelity since the changes over the weekend.


jmagly:
the overly helpful merge tool brought over more than expected. i will correct.

jmagly:
ok corrected, dont do deployment forks on github often.

Issue #343, opened 2023-04-01
Missing server response header


seanchito:
Firefox doesn't stream the server response correctly -- the response buffers for a few seconds, and then begins to stream.  

I noticed there is no `content-type` header for the streaming response.   Looking at Openai's ChatGPT version, which works in Firefox, the `content-type` is set `content-type: text/event-stream; charset=utf-8`  

So I modified api/chat.ts like so:

`return new Response(stream, { headers: { 'Content-Type': 'text/event-stream; charset=utf-8' } });`

And now everything works.   Plain text response from the server doesn't tell the browser what to do with the content, so bad things might happen in some browsers.

PR #700, opened 2023-05-04
fix streaming content-type header for firefox

jpetrucciani:
This PR adds the correct content-type to allow firefox to stream responses as expected. Thanks #343 for figuring the issue out!


mauhackr (approved):

mr-karan:
Would be great to have this soon! :)

PR #837, opened 2023-06-18
Llama

Romain-Boudot:


nikhilPatilGit:
why was this closed Romain ?

Romain-Boudot:
> why was this closed Romain ?

Don’t work properly for now and don’t have time to keep working on it, don’t even know if it’s the right way to do it too.
But i could have just left it like this i don’t know

PR #838, opened 2023-06-19
Add IDE-extension for `i18n` 

NilsJacobsen:
> Feel free to propose what needs to be changed in order to merge this PR

I added the `inlang` [IDE extension](https://inlang.com/documentation/apps/ide-extension) to chatbot-ui. That should make the life of devs easier while developing (extracting and inline annotations). I added the english translations to the `en` locales folder, in order to let the extension access the default language. 

#### Extract Messages (translations)

Extract Messages (translations) via the `Inlang: Extract Message` code action.

Extract Messages

#### Inline Annotations

See translations directly in your code. No more back-and-forth looking into the translation files themselves.

Inline Annotations


NilsJacobsen:
Besides this PR. Is there a reason why u decided to use the english translation as key? 
I guess it is hard to edit englisch translations and it seems to be complicated to add a new translation. `i18next` as far as I understood the library i would add the new translation key and string to englisch and then when making the translations you u just add the key with string to an other file. Just for clarification, just curious:)

PR #847, opened 2023-06-21
Add gpt-3.5-turbo-16k model

fre2d0m:


types/openai.ts
```diff
@@ -42,4 +43,10 @@ export const OpenAIModels: Record<OpenAIModelID, OpenAIModel> = {
     maxLength: 96000,
     tokenLimit: 32000,
   },
+  [OpenAIModelID.GPT_3_5_16K]: {
+    id: OpenAIModelID.GPT_3_5_16K,
+    name: 'GPT-3.5-16K',
+    maxLength: 96000,
+    tokenLimit: 32000,
+  }
 };
```
iam-Akshat, line 52: token limit is 16k no?

benedikt90: yes

PR #854, opened 2023-06-24
fixed bug of google search plugin so that it could function for most …

rogerzhuang:
the bug was caused by a typo:
value: updateConversation
where it should be:
value: updatedConversation


swiftdev29:
It still doesn’t work for me on Vercel…. I never get returned a response.

rogerzhuang:
i just fixed a typo bug. it would now return answers for some well structured questions. but functionality wise, some poorly structured questions may still not work due to search results and respective pages of the results exceeding max allowed tokens.

andresleecom:
Issue fixed on google search, thanks.

PR #904, opened 2023-07-16
fix: resolve next js build time environment variables from .env.local

### LINKED ISSUE #535: Some environment variables from compose are ignored
Hi,

I try to edit ``NEXT_PUBLIC_DEFAULT_TEMPERATURE`` and ``NEXT_PUBLIC_DEFAULT_SYSTEM_PROMPT`` (as renamed in [this commit](https://github.com/mckaywrigley/chatbot-ui/commit/ea1d09244acc3c94a87a3658b1ee271542c6fecf#diff-110077186eda23f5a76bd65631d9689184e89b97b96806abc2b05d15c8e42624)) in compose.yaml : 

```yaml
services:
  chat:
    container_name: chatbot-ui
    image: ghcr.io/mckaywrigley/chatbot-ui:main
    environment:
      TZ: 'Europe/Paris'
      OPENAI_API_KEY: "${OPENAI_API_KEY}"
      OPENAI_ORGANIZATION: MYORG
      NEXT_PUBLIC_DEFAULT_TEMPERATURE: 0.5
      NEXT_PUBLIC_DEFAULT_SYSTEM_PROMPT: Answer as if you were a duck
```

The ``OPENAI_API_KEY`` variable is correctly read, but ``NEXT_PUBLIC_DEFAULT_TEMPERATURE`` and ``NEXT_PUBLIC_DEFAULT_SYSTEM_PROMPT`` are ignored. 

I thinks it was supposed to be fixed by this issue https://github.com/mckaywrigley/chatbot-ui/issues/388 but it seems it's still broken.

Thanks for the good work anyway.

bartolkaruza:
This is related to #388 and I think also fixes #535.

The issue with next js build not having the environment variables is caused by how next js looks up environment variables. It requires a .env.local to be present, which is not present in the docker build step. The reason is that it's included in the .dockerignore, so the context doesn't have a .env.local.

This change creates an explicit .env.local created for the next build (under npm run build).

It uses docker secrets to make sure none of the other secrets leak into the cached images.


Eric-Catalyst:
Does not resolve 535.

PR #938, opened 2023-08-16
Add new localization - Traditional Chinese (zh-TW)

pichuang:
Provide Traditional Chinese (zh-TW) i18n


pichuang:
Based on this Issue https://github.com/i18next/i18next/issues/1467#issuecomment-716260436 dealing with `zh-TW`

PR #985, opened 2023-09-16
adding persian translation files

JavidChaji:
add persian (farsi) (fa) to the public locale


PeterDaveHello (commented):
You'll also need to update the locale list in `next-i18next.config.js` file.

JavidChaji:
> You'll also need to update the locale list in `next-i18next.config.js` file.

Thank you so much for mentioning that, I almost Forgot it.

PR #998, opened 2023-10-12
Thirtytwok

brandco:
migrating to API version 2023-09-01-preview, fixing issue for 32k model, adding 16k model, bug fixes, adding app insights telemetry

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN npm ci
 
 # ---- Build ----
 FROM dependencies AS build
+ARG NEXT_PUBLIC_APP_INSIGHT
 COPY . .
 RUN npm run build
 
--- a/components/AzureAppInsights/AzureAppInsights.tsx
+++ b/components/AzureAppInsights/AzureAppInsights.tsx
@@ -1,0 +1,37 @@
+import React, { ReactNode } from "react";
+import { ApplicationInsights } from "@microsoft/applicationinsights-web";
+import {
+    ReactPlugin,
+    withAITracking,
+    AppInsightsContext,
+} from "@microsoft/applicationinsights-react-js";
+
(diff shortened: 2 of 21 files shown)


brandco:
sorry, this PR was intended for my fork

Kush-agra-wal:
Hey can you tell me how you migrated to 2023-09-01

brandco:
It's in utils/server/index.ts at the end of this [diff ](https://github.com/mckaywrigley/chatbot-ui/pull/998/commits/d1bef50647cc519e5c37c23630cbabfde8388257)

The test for `finish_reason` was previously enough to find the last chunk of data in the stream from the API but that stopped working. So, the server was erroring out before the response was complete, ending the stream to the client and cutting short the response that users would see in the app. The streaming API now returns `"[DONE]"` when the stream is done. So I added an early return statement when `data === "[DONE]"`


          try {

            if (data === "[DONE]") {
              return;
            }
            const json = JSON.parse(data);
            if (json.object == "") {
              return;
            }
            if (json.choices[0].finish_reason != null) {
              controller.close();
              return;
            }
            if (json.object == "chat.completion.chunk") {
              const text = json.choices[0].delta.content;
              const queue = encoder.encode(text);
              controller.enqueue(queue);
            }
          } catch (e) {
            controller.error(e);
          }

          
         
You also need to set the OPENAI_API_VERSION environment variable.

Kush-agra-wal:
I guess azure removed the /deployments endpoint to give list of deployments , so you hardcoded the list? and also it seems your deployment id name =model name?

brandco:
Yes that's correct. I found that Azure has another endpoint for model deployments in an API for Azure resources associated with the Azure account. I was in a hurry and decided not to use it.

PR #1008, opened 2023-10-25
Azure specific changes

jtdevops:
- Fixing an issue with the Azure GPT-4 instances where the Azure Models API endpoint '/openai/deployments' responds with a 404 'Resource Not Found' error for newer GPT environments. Providing the ability to change 'deployments' to 'models' to fix this issue with a new environment variable AZURE_MODELS_PATH.
- When starting a new chat, the models drop-down can show the model name and ID to help distinguish between multiple models having the same name.
- Providing the ability to only show 1 model in the drop-down when the AZURE_DEPLOYMENT_ID environment variable is provided.
- If AZURE_DEPLOYMENT_ID is not provided, the models drop-down will show all available models, and use the selected model for the Chat Completion API requests.
- Fixed an issue where Azure provides a '[DONE]' response which breaks the JSON.parse function and where the JSON response doesn't contain any 'choices'. Each of these cause the UI to break.
- Ignoring '.env' files from being committed
- Updated the README.md with the changes mentioned above that affect the environment variables.


pureit-dev:
I am hitting this exact issue.  When will these changes be merged? Or is there a short term workaround?

ephb:
@pureit-dev 
I had the same issue and this PR solves it. A short term workaround is to build the docker image yourself and run it locally or push it to GitHub or Azure container registry to deploy it. I am using a private registry so I can not share.

PR #1051, opened 2024-01-09
Add one click install on sealos

fanux:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,10 @@ You will also need to to fill in the values for project_url
 npm run chat
 ```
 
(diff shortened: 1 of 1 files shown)


mckaywrigley:
Looks interesting but we're going to be a bit picky with 3rd parties.

fanux:
@mckaywrigley [Sealos](https://github.com/labring/sealos) is also an open-source project with 12k stars and is one of the largest container hosting platforms in China. It can bring Chatbot-UI to hundreds of thousands of Chinese users. I'm not sure about the specific requirements for third-party platforms.

PR #1064, opened 2024-01-09
Add Docker prerequisite

thenameless7741:
This PR updates the README to mention Docker as a prerequisite for running Supabase locally.


mckaywrigley:
Thanks for bringing this up! We added this change somewhere between you creating this and me seeing it :)

Issue #1091, opened 2024-01-10
Override API keys globally for users


jsweetzer-ea:
I would like to be able to override the API key for each provider by environment variable, such that all users are forced to use the provided API key. It would also be beneficial to then disable bring-your-own-key for any providers not enabled at env level. 

Is this something you would support adding in the short term?

housser-apply:
Agreed, we'd love to be able to deploy this for an internal team, and want all users to be using our centrally managed API keys without having to provision and distribute separate keys to all users.

sebiweise:
It would be nice to share a workspace and then define api keys for that workspace so every user that is working in that workspace will use the specific api key

mckaywrigley:
We’ll add support for this soon!

jsweetzer-ea:
Great, thank you!

tomasmcm:
Plus one on this. We use chatbot-ui with azure openai to provide a private instance of gpt-4 for internal use within the company and would like to abstract API keys like in V1

Issue #1066, opened 2024-01-09
Feature Request | API Key via env variable


kyleroberts-ai:
Hi,

Thank you for the amazing work.

I was wondering if instead of individual users uploading an API key, instead could we have a method to add via an env variable? 

Thanks,

mckaywrigley:
Now that the database is secure (not local storage) we really want to encourage people to use the UI as it's a bit more simple for non-technical users. Highly encourage you to clone the repo and implement it for yourself though. We want to encourage people to experiment with their own features :)

PR #1093, opened 2024-01-10
Store API keys in env to enable preset keys

tim13246879:
#1066 #1091 
Support the option to include API keys as envrionment variables. If a key exists in env, it overwrites the key provided by the user. Otherwise, if a key doesn't exist in env, the user-provided key is used.


Bortus-AI (approved):

mckaywrigley:
I think this is definitely something people want. Good start here but I'll probably need to go in and handle some conditional UI stuff as well.

im-calvin:
What UI stuff were you thinking about? I'm interested in contributing to this PR myself

wrapss:
I think we need an admin section

mckaywrigley:
@im-calvin Thinking something along the lines of hiding api key input in profile settings for providers that have api keys in the .env file, and then setting that key on the profile.

im-calvin:
Perfect, that's exactly what I was working on. I'll commit to the fork when I'm done (today/tomorrow)

Also looking into unlocking the menu for modifications for selecting the LLM in the top right

.env.local.example
```diff
@@ -7,3 +7,10 @@ SUPABASE_SERVICE_ROLE_KEY=
 
 # Ollama
 NEXT_PUBLIC_OLLAMA_URL=http://localhost:11434 # Default
+
+# API Keys (Optional: entering an API key here allows users to chat without entering their own key)
+OPENAI_API_KEY=
```
Catyousha, line 12: dont forget to add `NEXT_` prefix so the envs could be detected in client side.

im-calvin:
Made some changes for the API keys for user settings, made it so that if the environment variable is set, then it will disable the input in the user settings (so they can't set the API keys). Each of the fields can be modified on a per-field basis: if the environment variable is set for `OPENAI_API_KEY`, then it'll disable the `<Input />` field for that one, but the rest of the ones will be inputtable.

.env.local.example
im-calvin: I think this could be unsafe [(src)](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser)? I made another route in the API to fetch if a key exists. see 1e67b90cabecb45b58192b43b13ebae750c20a96

app/api/retrieval/keys/route.ts
```diff
@@ -0,0 +1,53 @@
...
+export enum VALID_KEYS {
+  OPENAI_API_KEY = "OPENAI_API_KEY",
+  ANTHROPIC_API_KEY = "ANTHROPIC_API_KEY",
+  GOOGLE_GEMINI_API_KEY = "GOOGLE_GEMINI_API_KEY",
+  MISTRAL_API_KEY = "MISTRAL_API_KEY",
+  PERPLEXITY_API_KEY = "PERPLEXITY_API_KEY",
+  AZURE_OPENAI_API_KEY = "AZURE_OPENAI_API_KEY"
+}
+
+export async function POST(request: Request) {
```
babarkamal114, line 6: I believe the logic can be enhanced further by incorporating reusable functions.

e.g: 
```
function createResponse(data, status): Response {
  return new Response(JSON.stringify(data), {
    status,
    headers: {
      "Content-Type": "application/json"
    }
  });
}
```

babarkamal114: Usage 

```
 if (!key) {
      return createResponse({ error: "Key type is required" }, 400);
    }
```

im-calvin: done in b4ea508

mckaywrigley:
Really nice work here everyone! Made a slight tweak to change the UI in the profile settings.

Everything works as far as I could tell. Awesome addition! Merging.

Issue #1047, opened 2024-01-04
Don't wait any longer; the project release has been postponed again. It's really frustrating; I've been waiting for a whole week already.


hikafeng:
![image](https://github.com/mckaywrigley/chatbot-ui/assets/76583915/02627c7a-9c53-4e62-bfdc-ec5810b947c9)
A little under the weather and would like to be 100% for launch as it’ll be fairly demanding, so pushing back a couple days. Friday or Monday - probably the latter. Everything is done though.

xiaos911:
Very regrettable

maddes8cht:
Well, it's ready when it's ready.
However, for the waiting people it's really frustrating to HAVE something like a "deadline" that gets postponed several times at the very moment when the release was scheduled.
If possible, postpone earlier.
Make more reliable anouncements.
And maybe, if people are waiting for it:
Make an alpha release or a beta release so people have  at least ANYTHING.
Things can always be improved.
There is no such thing as a "perfect" release.

vmldude:
Well said maddes8cht

twf-nikhila:
The fact that this project is open source & free is truly remarkable. I think instead of criticising we should respect @mckaywrigley for making it open-source

maddes8cht:
@twf-nikhila I fully agree, and i forgot to mention that.
Yes, always, i should have also added a BIG and LOUD "Thank you" for the great work !

Anyways, there is always room for improvements, and exposing oneself with a project like this always includes the option to give constructive criticism.
Given that:
At this point, i would have loved to get an alpha- or beta release mentioning the current problems that still needs to be resolved.

Again:
Love this project.
Thanks for the great Work.
I'm caring for it, so i'm eagerly waiting for the rlease.

enricoros:
You should check out other projects as well, such as https://github.com/enricoros/big-agi :)

MrBenzWorld:
> You should check out other projects as well, such as https://github.com/enricoros/big-agi :)

Thank you, it is great . I was using llama models for months and I used all which are mentioned  by #thebloke,but I did not know about it. This is the best chat ui with a clean and clear interface. 
@enricoros

twf-nikhila:
This looks amazing, thanks for sharing it @enricoros

jonschlinkert:
[...]

PR #1096, opened 2024-01-11
Improve placeholder text in chatbox to mention @ and / commands

tim13246879:
This PR tells users how to refer to uploaded files and prompts in a workspace using @ and / commands. 

It took quite a while to figure out how to access uploaded files in user queries. Before stumbling on the twitter post at #1047 I had to manually add the same file for each new chat :(

Since this is a useful feature, it would be nice to have a bit more guidance for new users. Screenshots below show a Before Vs After comparison. The only difference is in the text in the chatbox.

image

image


chadrosser (approved):

mckaywrigley:
Made a minor change and merged. Welcome to contributors list!

PR #1101, opened 2024-01-11
feat: fix scroll hook

Lantianyou:


mckaywrigley:
I see where you're coming from here but the intended behavior right now is that you *should* be able to cancel the autoscroll which this reverts.

PR #1142, opened 2024-01-13
add file navigation with arrow keys

im-calvin:
feat: allows users to press up arrow / down arrow navigation to select file without moving the mouse


mckaywrigley:
Works great! Thanks for the add. Really nice work.

Issue #1072, opened 2024-01-09
Ollama not working


Bortus-AI:
I tried the latest ollama commit but still can't get Ollama models to show up in chatbot. 

I tried two different setups

1. Debian server with ollama and chatbot-ui 2.0 running locally
2. Vercel deployment and setting env to point to my ollama server that accepts remote connection.

Ollama server is confirmed working when using unsaged, ollama-webui and big-AGI

I am unable to see any attempts from chatbot-ui 2.0 to ollama and nothing is showing in the local server logs or in vercel. 

<img width="461" alt="image" src="https://github.com/mckaywrigley/chatbot-ui/assets/100248124/586c1a48-dfae-434f-b3e6-1b4d959e899b">

wrapss:
see [#1062](https://github.com/mckaywrigley/chatbot-ui/pull/1062)

Bortus-AI:
> see [#1062](https://github.com/mckaywrigley/chatbot-ui/pull/1062)

That fixed it! Thanks so much

![image](https://github.com/mckaywrigley/chatbot-ui/assets/100248124/30069b3b-b369-4e6a-be7c-48b51cbbb1c6)

Bortus-AI:
@wrapss 
Models show up now but when trying to chat with them I get 

![image](https://github.com/mckaywrigley/chatbot-ui/assets/100248124/7e3262f6-ec7a-4343-8f62-d46e2960b0fe)

wrapss:
u launch ollama with arg  OLLAMA_ORIGINS=*  ?

wrapss:
> @wrapss Models show up now but when trying to chat with them I get

u launch ollama with arg OLLAMA_ORIGINS=* ?

Bortus-AI:
> > @wrapss Models show up now but when trying to chat with them I get
> 
> u launch ollama with arg OLLAMA_ORIGINS=* ?

Yes I did. Interesting enough its working with Vercel but not my local instance. Let me double check the file changes to make sure I didn't have a typo

Bortus-AI:
Well it works with Vercel so at least your fix is working for that. I will redo my local setup after lunch and report back. 

Thanks for getting the models to work!

Bortus-AI:
Seems to be no longer working on Vercel anymore. Not getting models to show up now

Bortus-AI:
[...]

PR #1146, opened 2024-01-13
Fetch Ollama models on client side

bernie43:
This PR makes it possible to use local Ollama models on a hosted version of chatbot-ui.

Previously the request to fetch the list of models from the local Ollama server was proxied via a Next.js endpoint, but the requests later in the actual chat are already done directly to the local server.
This proxying seemed unnecessary to me anyway.

See also my comment here: https://github.com/mckaywrigley/chatbot-ui/issues/1072#issuecomment-1889868387


mckaywrigley:
As far as I can tell you're spot on here. Good call!

PR #1149, opened 2024-01-13
Issue#1128: Disabled signup by setting enable_signup = true to [auth]…

imanpatelb:
config.toml
setting enable_signup = true to [auth], [auth.email], [auth.sms]


mckaywrigley:
By default we want to keep this on. If you want to disable it for specific project totally makes sense though!

PR #1152, opened 2024-01-13
Redirect to /chat on index if user is logged in

Mikodin:


middleware.ts
```diff
@@ -5,7 +5,13 @@ export async function middleware(request: NextRequest) {
   try {
     const { supabase, response } = createClient(request)
 
-    await supabase.auth.getSession()
+    const session = await supabase.auth.getSession()
```
jollytoad, line 8: @mckaywrigley this always returns a value, the actual session is nested within an object, it should probably be:

```
const { data: { session } } = await supabase.auth.getSession()
```

PR #1153, opened 2024-01-13
Adds .nvmrc for Node 20.11.0 LTS

Mikodin:


mckaywrigley:
Is there a particular reason why this is needed? I'm clueless here tbh.

Mikodin:
> Is there a particular reason why this is needed? I'm clueless here tbh.

No worries - it ensures that people are on a consistent version of node.  Really good for people working across multiple projects with varying versions of node.

It's common across many popular projects
https://github.com/facebook/react/blob/main/.nvmrc

mckaywrigley:
Got it. Seems smart and thanks for the info.

PR #1155, opened 2024-01-13
Add Gitpod and Devcontainer setup

K-Kit:
Saw people complaining about difficult setup and thought this might be helpful.

It would be nice to get everything working in a browser IDE but I was unable to find a clean way to do so due to security settings.


README.md
```diff
@@ -26,6 +26,10 @@ Chatbot UI was recently updated to its 2.0 version.
 
 The code for 1.0 can be found on the `legacy` branch.
 
+## 1 Click Quick Start
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/k-kit/chatbot-ui)
```
K-Kit, line 31: If merged this needs to change to main repo url.

mckaywrigley:
Can you give me a TLDR on the what/why of this. I'm admittedly a bit unfamiliar - thanks!

K-Kit:
> Can you give me a TLDR on the what/why of this. I'm admittedly a bit unfamiliar - thanks!

Totally, the main reason for the PR is ease of first time setup.  Im going to use gitpod in examples for simplicity but devcontainers can provide similar benefits just locally and require docker to be installed.

The button should prompt you to create a gitpod instance which is basically a remote vm. 
By default it already has an optimal Linux environment for development with docker, node, brew, etc included.
When the pod is initialized it runs the setup script automatically.
With the consistent environment the script should succeed every time and automatically start all the dev processes.  The script even sets up your environment variables for you.

So you want to set up a fresh environment you launch a pod, go make a coffee and your good to go.

This means even the kid who just saw your YouTube video, maybe can write a hello world but has never seen npm, docker etc.  can open up a code editor and start playing with it in under 10 minutes 

Then in a more general sense there's a number of advantages to using remote dev containers.  For me the biggest thing is not having to mess with my own env when checking out my teams PRs and not having to waste time helping people get their env right when they're using lesser operating systems(windows).

I could go on but this is starting to feel like an ad. Let me know if you have any specific questions or I can clarify something.

Edit: the dev container file is also used for GitHub code spaces but I was having even more trouble with the cross origin requests because it wouldn't forward to localhost.  If we can make the supabase api url an env variable somehow that might be a slightly more beginner friendly option.

K-Kit:
> Can you give me a TLDR on the what/why of this. I'm admittedly a bit unfamiliar - thanks!

this happened to pop up on my feed about dev containers today.
https://www.youtube.com/watch?v=dihfA7Ol6Mw

mckaywrigley:
I'm going to put this on hold for now but may revisit later.

Merged the OpenRouter one though which was exceptional!!

PR #1163, opened 2024-01-14
New provider: Open Router

K-Kit:
Adds openrouter for easy access to a variety of models
[https://openrouter.ai/rankings](https://openrouter.ai/rankings)
image


mckaywrigley:
Very in-favor of supporting Open Router. I'll review this later today. Definitely something I'd like to support!

app/api/chat/openrouter/route.ts
```diff
@@ -0,0 +1,47 @@
+import { CHAT_SETTING_LIMITS } from "@/lib/chat-setting-limits"
+import { checkApiKey, getServerProfile } from "@/lib/server/server-chat-helpers"
+import { ChatSettings } from "@/types"
+import { OpenAIStream, StreamingTextResponse } from "ai"
+import OpenAI from "openai"
+import { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions.mjs"
+
+export const runtime = "edge"
```
babarkamal114: export const runtime: string = "edge";

app/api/chat/openrouter/route.ts
```diff
@@ -0,0 +1,47 @@
...
+      messages: messages as ChatCompletionCreateParamsBase["messages"],
+      temperature: chatSettings.temperature,
+      max_tokens: undefined,
+      stream: true
+    })
+
+    const stream = OpenAIStream(response)
+
+    return new StreamingTextResponse(stream)
+  } catch (error: any) {
```
babarkamal114: error: Error

app/api/chat/openrouter/route.ts
```diff
@@ -0,0 +1,47 @@
...
+    const errorMessage = error.error?.message || "An unexpected error occurred"
```
babarkamal114: error.error?.message over  error.message why this approach is used for error handling.

babarkamal114: Instead of hardcoding HTTP status codes like 500, introduce a Status constant object for better code readability and maintainability. 
e.g
const Status = {
  SERVER_ERROR: 500,
};

const errorCode = error.status || Status. SERVER_ERROR

components/ui/chat-settings-form.tsx
```diff
@@ -100,12 +100,22 @@ const AdvancedContent: FC<AdvancedContentProps> = ({
...
 }) => {
-  const { profile, selectedWorkspace } = useContext(ChatbotUIContext)
+  const {
+    profile,
+    selectedWorkspace,
+    availableOpenRouterModels,
+    selectedAssistant
+  } = useContext(ChatbotUIContext)
+
+  function findOpenRouterModel(modelId: string): any {
```
babarkamal114: I noticed that you are using the type 'any.' Is it possible for you to implement specific types for it?

interface OpenRouterModel {
  modelId: string;
}

function findOpenRouterModel(modelId: string): OpenRouterModel | undefined

components/utility/global-state.tsx
```diff
@@ -271,6 +276,36 @@ export const GlobalState: FC<GlobalStateProps> = ({ children }) => {
...
+    try {
+      const response = await fetch("https://openrouter.ai/api/v1/models")
+
+      if (!response.ok) {
+        throw new Error(`Open Router server is not responding.`)
+      }
+
+      const { data } = await response.json()
+
+      const openRouterModels = data.map((model: any) => ({
```
babarkamal114: Same here

babarkamal114 (commented):
Overall, the pull request looks good, I left some minor comments. I noticed that you are using the type any in many places. Is it possible for you to implement specific types for it?

K-Kit:
> Overall, the pull request looks good, I left some minor comments. I noticed that you are using the type any in many places. Is it possible for you to implement specific types for it?

Thanks for the review.  Sorry it was a bit sloppy I don't know why I felt the need to make a pr before going to bed at 6am.  Intrusive thoughts won that battle.

Yes definitely I will make those changes this tonight.  

If I recall correctly the route endpoint was an exact copy of the openai route with just one line added for the baseURL. 
Would it be preferable to just extend that instead?

K-Kit (commented):
OpenRouter typing added

app/api/chat/openrouter/route.ts
K-Kit: Added server runtime type. its actually a union of 3 string literals and undefined :P

app/api/chat/openrouter/route.ts
K-Kit: This is actually copied from the openai endpoint I didn't even look at it.  That being said it might be worth refactoring both with your suggestions and abstracting it out into an openAIapiHandler function thats shared to keep things DRY.  I don't mind doing that but should it be in this PR or another one?

components/ui/chat-settings-form.tsx
K-Kit: Done

app/api/chat/openrouter/route.ts
K-Kit: Added an ErrorResponseSchema to parse the error and ensure proper typing

PR #1166, opened 2024-01-14
Fix Process Docx Bug

hikafeng:
add new function
createFileBasedOnExtension


mckaywrigley:
What bug does this fix?

There is a `createDocXFile` funciton that should be handling the docx files.

hikafeng:
You can try to upload, you will find that uploading DOCX files will fail, the reason is that some functions are missing.

hikafeng:
The reason is that the unmodified components/sidebar/items/all/sidebar-create-item.tsx file directly calls the const createdFile = await createFile function, which is used to process non-DOCX files. When uploading a DOCX file, calling createFile will obviously fail because it is not designed to handle DOCX files.

mckaywrigley:
Got it. Will merge and nice catch.

PR #1177, opened 2024-01-15
Feature/#1110 dall e

Chr96er:
Warning: The PR is not ready to be merged.

OpenAI chats use the DALL-E API to generate images if a prompt starts with "Generate an image". It then uses the completions API to make the prompt more elaborate before sending the request to the image generation API.

Currently, the function only returns the image URL so it doesn't render in the UI yet. @mckaywrigley would you be able to point me in the right direction for making the images render in the UI? You can see some of my failed attempts in the [prior commit](https://github.com/mckaywrigley/chatbot-ui/commit/3247f09c854fd7b80e601571260f171641c314ae). Many thanks!

Also, I think the image parameters (DALL-E 2 vs 3; HD vs standard; resolution) should be provided in the settings.


mckaywrigley:
@Chr96er We're going to be implementing DALL-E with our tools update later this week.

Right now the images code absolutely sucks so refactoring that will be part of it.

I suspect if you wait 3-4 days this will all be taken care of, so you may just want to hang tight and check back then!

Chr96er:
Amazing, thanks @mckaywrigley!

erahhal:
Did the work for this feature get moved to a different PR?

Chr96er:
@erahhal No, it was rejected because the images code was supposed to change that week. I haven't kept exact track of what happened since but I think dall-e hasn't been integrated yet. I'm using my fork at the moment, because I only require image generation. It works (returning the dall-e image url), but I agree that it shouldn't be merged as it is right now.

Issue #1180, opened 2024-01-15
Bug: Japanese typing conversion issue


gijigae:
Unlike English, Japanese requires conversion from Hiragana to Kanji and this conversion is often triggered by the enter key. Thus, while typing messages in Japanese, `handleKeyDown` needs to consider whether the user is still composing text. Current implementation sends the text right after the enter key is pressed. Instead, messages should be only sent after the text is fully composed.

![image](https://github.com/mckaywrigley/chatbot-ui/assets/81459/14788b1f-1632-4c18-9138-d56c73c73208)

mckaywrigley:
Merged.

ozora-ogino:
@mckaywrigley @gijigae Thank you for fixing this bug, but the same bug happens when I typing japanese text in other text box (e.g. create prompt or preset) . I would like you to apply this changes to all text boxes.

gijigae:
@ozora-ogino Thank you so much for the kind words and suggestion. Made necessary changes and currntly running a few tests. If everything goes well, I'll most likely create a PR tomorrow. Stay tuned.

ozora-ogino:
@gijigae Thank you so much🙏

gijigae:
@ozora-ogino Created #1203 to cover workflows for Prompts. Will raise another PR to fix other workflows.

PR #1181, opened 2024-01-15
Fix Japanese typing issue

gijigae:
This PR fixes the issue reported in #1180 .


mckaywrigley:
Awesome. Thanks for doing this!

PR #1183, opened 2024-01-15
Hides sidebar when clicking outside of its area

Mikodin:


mckaywrigley:
Sidebar should always stay open unless toggled.

Issue #1192, opened 2024-01-16
PDF upload not working for Azure OpenAI


othmanelhoufi:
Hello,

Thanks for this great tool, it is indeed the best Chat UI I've tried so far!

Let's get to the problem at hand. When uploading a file the default embedding model used is only from OpenAI services even though I am connected to the Azure OpenAI services.

Here is the error:
```
Error: OpenAI API Key not found
    at checkApiKey (webpack-internal:///(rsc)/./lib/server/server-chat-helpers.ts:52:15)
    at POST (webpack-internal:///(rsc)/./app/api/retrieval/process/route.ts:31:89)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/ohoufi/Projects/chatbot-ui/node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js:6:62591
 ✓ Compiled /chat/[chatid] in 437ms (4547 modules)
Error: OpenAI API Key not found
    at checkApiKey (webpack-internal:///(rsc)/./lib/server/server-chat-helpers.ts:52:15)
    at POST (webpack-internal:///(rsc)/./app/api/retrieval/process/route.ts:31:89)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/ohoufi/Projects/chatbot-ui/node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js:6:62591
Error: OpenAI API Key not found
    at checkApiKey (webpack-internal:///(rsc)/./lib/server/server-chat-helpers.ts:52:15)
    at POST (webpack-internal:///(rsc)/./app/api/retrieval/process/route.ts:31:89)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/ohoufi/Projects/chatbot-ui/node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js:6:62591
```

Looking down into the `route.ts` file located in the `retrieval` folder I can see that the logic behind sends requests to OpenAI or Local embedding models and not to the currently connected provider:
```
if (embeddingsProvider === "openai") {
      checkApiKey(profile.openai_api_key, "OpenAI")
    }

    let chunks: any[] = []

    if (embeddingsProvider === "openai") {
      const openai = new OpenAI({
        apiKey: profile.openai_api_key || "",
        organization: profile.openai_organization_id
      })

      const response = await openai.embeddings.create({
[...]

Issue #1103, opened 2024-01-11
Azure OpenAI embeddings currently not supported


Enkronan:
Only embedding options are currently _local_ or _openai_, meaning even though I've selected Azure OpenAI as my preference, when uploading files it'll break unless I have also specified an openai_api_key.

mckaywrigley:
Now supported!

PR #1197, opened 2024-01-17
Add support for Azure embeddings provider

dino1729:
This will fix issue #1192, #1103.

The only assumption is DEPLOYMENT_ID = "text-embedding-ada-002". If necessary, UI element to capture the deployment id from the .env file of the UI like the other azure LLM deployment IDs can be added.


mckaywrigley:
For anyone seeing this I will merge tonight (just need to change a few minor things).

Apologies for this bug, and thanks for a great PR!

mckaywrigley:
Everything looked great - I just made one architecture change.

Since both OpenAI/AzureOAI both serve text-embeddings-ada-002 I'm not separating how they're stored in the DB.

PR #1219, opened 2024-01-20
Enable ollama multimodal 

wrapss:
Enable ollama multimodal (only images in reality)
Open to a better solution to find out if a model is supplied by ollama or not


mckaywrigley:
Definitely want to enable this.

One consideration is that on the frontend we restrict image input to specific models.

Is there a way to tell if an Ollama model allows image input via the local Ollama API?

wrapss:
in part yes, [see](https://github.com/jmorganca/ollama/issues/1772)

PR #1221, opened 2024-01-21
Feature: Email whitelist

francofantini:
### Description:

This PR introduces optional additional validation for the signup process by checking the domain and/or specific email addresses against a whitelist. Users attempting to sign up with an email domain or email address that is not included in the whitelist will be redirected to the login page with an appropriate error message.

### Changes:

Added a check for email domains against a whitelist defined in the environment variable `EMAIL_DOMAIN_WHITELIST`. If the user's email domain is not in the whitelist, the signup process is halted, and the user is redirected with an error message.
Implemented a similar check for specific email addresses using the environment variable `EMAIL_WHITELIST`. If the user's email is not in this whitelist, they receive a redirect with an error message.
Both whitelists are optional; if the respective environment variable is not set or is empty, the check is bypassed, allowing any email to sign up.

### Motivation:

This feature was added to enhance security and control over who can register to the service. By allowing only certain domains and email addresses, we can ensure that only authorized users have access to our platform.


mckaywrigley:
love it!

VeckoTheGecko:
Just clarifying here a bit with the code, if `EMAIL_DOMAIN_WHITELIST` and `EMAIL_WHITELIST` are **both** set (for example `EMAIL_DOMAIN_WHITELIST="email@mycompany.com"` and `EMAIL_WHITELIST="example@gmail.com"`, wouldn't it be logical for `EMAIL_WHITELIST` to be checked first, and if failing then checking `EMAIL_DOMAIN_WHITELIST`? Or are these environment variables not meant to not be used together?

At the moment it looks like if `example@gmail.com` is used, then it will fail since its not a `mycompany.com` email.

PS: This is my first exposure to this codebase. I am by not able to submit a PR for this.

PR #1235, opened 2024-01-22
Fix: Excludes auth from middleware

francofantini:
This PR introduces a small yet crucial change to the middleware configuration. The modification ensures that requests to the /auth path are no longer processed by the middleware, which was previously causing issues with the password recovery flow.


mckaywrigley:
nice catch!

PR #1241, opened 2024-01-23
Add custom header support and tool-call parameters as request body for tools 

MartianInGreen:
Adds two things:

1. **Custom-Headers:** Adds support for custom headers for api calls (for authentication etc.)
2. **Request type:** Most api's want the request to be contained in the request body, this lets the user choose between the old way (in the url) and having the parameters of the tool call in the request body.


mckaywrigley:
Thanks for doing this. I'll get around to review tonight!

mckaywrigley:
Sorry got a little bogged down but this will be merged tonight for real haha

PR #1246, opened 2024-01-23
fix azure model name environment variable

### LINKED ISSUE #1245: Azure deployment set in environment variable
in `.env.local`, I set the `AZURE_GPT_45_VISION_NAME` to some value, but when I try and chat, `/api/chat/azure` throws an error: `Azure resource not found`. I see that `DEPLOYMENT_ID` in `/api/chat/azure/route.ts` is an empty string, and that `profile.azure_openai_45_turbo_id` is an empty string. 

I wonder if it's due to some variables being named `_name` and `_id` in the `.env` and `profile` 
#1207 might be relevant

im-calvin:
fixes #1245


mckaywrigley:
Thanks!!

PR #1247, opened 2024-01-24
fix origin not defined for email signup

im-calvin:
title. 

simple 1 line fix, origin wasn't defined for the signup, but is defined for the forget password


mckaywrigley:
Since email verification is turned off by default we don't need the origin in this function.

PR #1262, opened 2024-01-25
Added support for a CHATBOT_UI_PORT env var.

danlynn:
This is particularly helpful when running locally and you have other next.js or ruby on rails projects running that also use port 3000 by default.  If CHATBOT_UI_PORT env var is not provided then it defaults to port 3000.


mckaywrigley:
Agree that it's helpful, but it's a simple enough add that we'd prefer individuals users make that change on their own as we want to keep the env file uncluttered.

Issue #1283, opened 2024-01-26
file uploading failed on Netlify deployments


xycjscs:
I deploy the project on the Netlify. I can not upload files for embedding anymore. 
Local deployments are fine.

Anyone has clues?

xycjscs:
vercel is okay. However chinese\japaness\koreans sentence will be garbled.

Pirogiran:
Same issue here.
Edit: On Vercel not only Chinese (and others) will be corrupted, but also Russian, for example.

mckaywrigley:
Please move to "Help" in "Discussions" above

Issue #1171, opened 2024-01-14
[Bug] '��' in responses for Chinese/Japanese/Korean requests


haotian-liu:
Related to #1147, but it happens to pure chat queries as well. Solving this may solve #1147 as well.

> Reference prompts to reproduce (Chinese/Japanese/Korean)
> 写一篇1000字关于人类与AI的议论文
> 人間とAIに関する1000字の論文を書いてください。
> 인간과 AI에 관한 1000자 논의 글을 쓰세요.

<img width="723" alt="image" src="https://github.com/mckaywrigley/chatbot-ui/assets/6631389/453c0988-f56e-4faa-93a1-15e375d28c26">
<img width="731" alt="image" src="https://github.com/mckaywrigley/chatbot-ui/assets/6631389/7f54655c-b26a-46e0-a326-1635b80e1253">

gijigae:
Thanks for raising the issue! What is your setup for Chatbot UI? I have two environments; one is hosted in Vercel and the other is deployed locally. For me, the issue you reported only occurs in former.

haotian-liu:
@gijigae nice catch! it's on vercel (and with the online supabase). not sure which one is the cause.

gijigae:
My local environment is connected to the Supabase web service. Thus, the issue would be something to do with the way Vercel is hosted. I will try to deploy Chatbot UI into another hosting service and will report back in case the issue persists.

xycjscs:
Same problem on the Vercel. No idea how to debug.

Pirogiran:
On Netlify everything works fine, compared to Vercel.

xycjscs:
Moved project to Netlify, this problem avoided. 
I will keep looking on this issue, hope it could be solved.
some reference  https://github.com/ChatGPTNextWeb/ChatGPT-Next-Web/issues/507

mckaywrigley:
Combo of an OpenAI/Vercel issue.

xycjscs:
brothers, it is a bug. Problem has been solved. I will make a pull request quickly. 
Chinese/Japanese/Korean can keep following this project with Vercel deployment and avoid compatible problems at Netlify.

Issue #1147, opened 2024-01-13
【Bug】Uploading a docx file failed and when using Chinese dialogue, occasionally there are '��' in the answers.


ghost:
Uploading a docx file failed 
![image](https://github.com/mckaywrigley/chatbot-ui/assets/154501119/3d289f13-26a8-470f-bbe3-f39a8c51daa4)
![image](https://github.com/mckaywrigley/chatbot-ui/assets/154501119/195d553a-99d2-4a4e-bfa2-71944295b58e)

there are '��' in the answers.
![image](https://github.com/mckaywrigley/chatbot-ui/assets/154501119/726b38ae-8aa8-4767-8774-98d6b48b0595)

mckaywrigley:
I suspect that the Docx document loading method is decently buggy. It's admittedly a tougher file format to work with for RAG.

If anyone sees and has ideas PRs here would be awesome.

ghost:
According to the test, processing docx files actually calls \app\api\retrieval\process\route.ts and does not use \app\api\retrieval\process\docx\route.ts

mckaywrigley:
Try pulling? Should work. Made a type check change just in case that was the bug.

Subhahdip:
Hi guys!! just move into web dev recently. can you guys help me get my first contribution in chatbot-ui? really excited :)

hikafeng:
> Try pulling? Should work. Made a type check change just in case that was the bug.

[Fix Process Docx Bug](https://github.com/HikaFeng/chatbot-ui/commit/420dffc30349e14445fe1ac3db59684a2e20c4ac)

and  can you guys help me get my first contribution in chatbot-ui? really excited!!!!!!!!!!!!

PR #1296, opened 2024-01-27
Solve the garbled issue of Chinese/Japanese/Korean

xycjscs:
this pull request is to solve issues mentioned at https://github.com/mckaywrigley/chatbot-ui/issues/1171#issuecomment-1913188903

This is my first contribution. I am happy to share my excitement with you.

At first, we thought it may be resulted from the combo of Vercel and OpenAi. However, I checked the Vercel/Ai repo, and this problem has been address before. I deployed the openai example of Vercel/ai repo at vercel. It could deal with Chinese language pretty well. I believe it is a bug.

I dived into this project to find where the bug is. Characters are returned by chunk, so I add "1" between each chunk.
Screenshot 2024-01-28 at 12 06 41 AM
Clearly, the chunk is not complete, and the garble issue happens. 

Then, I found the decode function of each chunk  `callback(decoder.decode(value))` at the lib/consume-stream.ts
This function was mentioned in this issue. 
https://github.com/ChatGPTNextWeb/ChatGPT-Next-Web/issues/507

I believe, following correction can solve problem:
`callback(decoder.decode(value, { stream: true }))`

After testing. There is no more garbled issues.

Screenshot 2024-01-28 at 12 12 33 AM

People who deployed project at Netlify and files uploading issues happened, could keep using Vercel with this commit. 
https://github.com/mckaywrigley/chatbot-ui/issues/1283#issuecomment-1912350380

### DIFF
--- a/lib/consume-stream.ts
+++ b/lib/consume-stream.ts
@@ -17,7 +17,7 @@ export async function consumeReadableStream(
       }
 
       if (value) {
-        callback(decoder.decode(value))
+        callback(decoder.decode(value, { stream: true }))
       }
     }
   } catch (error) {


xycjscs:
？？ issue in https://github.com/mckaywrigley/chatbot-ui/issues/1147 could also be solved.

xycjscs:
Validations on Japanese/Korea
IMG_2160
IMG_2159

xycjscs:
Validation on Russia, in case others searched with Russia keyword 
IMG_2161

mckaywrigley:
You are a hero.

PR #1302, opened 2024-01-28
Update support for GPT-4-0125

hikafeng:


mckaywrigley:
gpt-4-turbo-preview defaults to that model already

PR #1322, opened 2024-01-29
Don't insert project ref in the command

Itangalo:
The description currently says that --project-ref should be changed to the project ID string. I'm pretty sure that's wrong.


mckaywrigley:
Good catch. Looks like this was changed.

PR #1334, opened 2024-01-30
Setup PWA

davecrab:
Added pwa components, have only been able to test in iOS


johan456789:
Does this mean this app can be saved as a PWA on mobile now?

PR #1363, opened 2024-02-02
Support advanced LLM from ZHIPU(ChatGLM): GLM-4 and GLM-3-TURBO.

Rid7:
Support advanced LLM from ZHIPU(ChatGLM): `GLM-4` and `GLM-3-TURBO`. For more details, see [ZHIPU AI API Documentation](https://open.bigmodel.cn/dev/api)


mckaywrigley:
Always love when people use other models, but this isn't a platform we want to support out of the box.

PR #1369, opened 2024-02-04
feat: add type for onScroll

Lantianyou:


mckaywrigley:
Always appreciate better typing - thanks!

PR #1374, opened 2024-02-04
Updated files.ts to encode filenames to base64

### LINKED ISSUE #1156: Bug: File upload using german umlaute as "name" doesn't work
Uploading a pdf file and giving a name containing german umlaute like "ö" doesnt work.
Using "oe" instead of "ö" works.

- ❌ works NOT: "ChatGPT_im_PM_MOVR_ManagementCircle_Köln" 

image

- ✅ WORKS: "ChatGPT_im_PM_MOVR_ManagementCircle_Koeln"

image

cheerlessDreamer:
Added base64 encoding to the file path to support characters outside of regular English.
This closes #1156

### DIFF
--- a/db/storage/files.ts
+++ b/db/storage/files.ts
@@ -15,7 +15,7 @@ export const uploadFile = async (
     throw new Error(`File must be less than ${SIZE_LIMIT / 1000000}MB`)
   }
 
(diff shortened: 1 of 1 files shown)


mckaywrigley:
Great solution!

PR #1378, opened 2024-02-05
Support multiple JSON objects in Ollama stream chunk

### LINKED ISSUE #1271: Local model output is terrible
I am using ollama to run local models the output through chat is just horrible but the output through ollama is great
Example below-

Both have the same prompt "Why is the sky blue?"

**ChatbotUI**
image

Vs **Ollama API**
```
The sky appears blue because of a phenomenon called Rayleigh scattering. When sunlight enters Earth's atmosphere, it encounters tiny molecules of gases
such as nitrogen and oxygen. These molecules scatter the light in all directions, but they scatter shorter (blue) wavelengths more than longer (red)
wavelengths. This is known as Rayleigh scattering.

As a result of this scattering, the blue light is dispersed throughout the atmosphere, giving the sky its blue color. The red light, on the other hand,
is able to travel longer distances without being scattered, which is why the sun appears red at sunset.

Other factors can also affect the appearance of the sky, such as the presence of aerosols (small particles in the air) or water vapor, which can
scatter light in different ways and change the apparent color of the sky. However, Rayleigh scattering is the main cause of the blue color of the sky
that we see.
```

zacharytamas:
In some cases Ollama can send more than one JSON object in a single chunk of the streamed response. These are separated with new-line characters and are thus not valid JSON as-is. This change splits the chunk by new-lines and parses them individually.

I believe this fixes #1271.

### DIFF
--- a/components/chat/chat-helpers/index.ts
+++ b/components/chat/chat-helpers/index.ts
@@ -298,7 +298,19 @@ export const processResponse = async (
         setToolInUse("none")
 
         try {
-          contentToAdd = isHosted ? chunk : JSON.parse(chunk).message.content
+          contentToAdd = isHosted
+            ? chunk
+            : // Ollama's streaming endpoint returns new-line separated JSON
+              // objects. A chunk may have more than one of these objects, so we
+              // need to split the chunk by new-lines and handle each one
+              // separately.
+              chunk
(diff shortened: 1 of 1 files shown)


mckaywrigley:
Super high value-fix!

PR #1385, opened 2024-02-06
fix: cannot automatically switch to available workspace after workspace is deleted

jzhangdev:


components/chat/chat-hooks/use-scroll.tsx
fkesheh, line 1: I would not do that in the same PR. Even if that is correct, which I'm not sure it's aligned with the prettier specs, it confuses the reviewer as he needs to go to other files. It also may pollute git blame. This in IMO

db/storage/files.ts
fkesheh, line 1: I would not do that in the same PR. Even if that is correct, which I'm not sure it's aligned with the prettier specs, it confuses the reviewer as he needs to go to other files. It also may pollute git blame. This in IMO

fkesheh (commented):
It looks good to me, but it has minor format changes on two other files that are not directly related to this PR

components/chat/chat-hooks/use-scroll.tsx
jzhangdev: Thanks for the review, I've removed the extraneous changes.

PR #1403, opened 2024-02-07
Add chat history navigation functionality

superhappychris:
Found myself wanting to reuse some or all of the text from a previous message in a lot my chats so I've added that ability by holding shift plus control plus the arrow keys when focused on the `ChatInput` component (similar to a command history in a shell). This will populate the textbox (no images or files come along).

I left the placeholder text alone, let me know if you want me to update that to something like: `Ask anything. Type "/" for prompts, "#" for files, "!" for tools, and shift+ctrl+arrow keys to navigate chat history.`


fkesheh (changes requested):
Please add the shortcuts on chat-help.tsx

Issue #1377, opened 2024-02-04
Getting Error at Sign up/Login


yesbhautik:
Logs:
![Termius_g41lmFeh4r](https://github.com/mckaywrigley/chatbot-ui/assets/63995315/b55419cf-1cb7-442f-9fa2-df5f45b3a073)

Error:
![chrome_dee2ibgjJQ](https://github.com/mckaywrigley/chatbot-ui/assets/63995315/0383e3c7-2319-47ba-93e1-d6f694d283b5)

tarek-kerbedj:
I'm getting the same error , found any solutions ?

jhchundev:
Have you migrated your supabase database?

superhappychris:
I ran into this problem running locally and it turned out there was some kind of port conflict with docker. To fix it I ran `supabase stop`, restarted my computer, then ran `supabase start`

yesbhautik:
I am using Local Method in my Ubuntu Server [because I dont wanna use vercel and supabase cloud]

yesbhautik:
> I'm getting the same error , found any solutions ?

I got the solution.

I have done as below & it works...

1 cloned repo
2 npm i
3 npm install supabase --save-dev
4 nano .env OR set values to env variable if you are using any panel for more security with [SUPABASE_ACCESS_TOKEN (supabase account api), NEXT_PUBLIC_SUPABASE_URL, NEXT_PUBLIC_SUPABASE_ANON_KEY, SUPABASE_SERVICE_ROLE_KEY, EMAIL_WHITELIST] 
5 modified "supabase/migrations/20240108234540_setup.sql"
6 supabase link --project-ref <project-id> [no need to enter any password]
7 supabase db push [need to enter password which you have set during database creation on supabase cloud]
8 npm run dev
-- DONE --

[if you are in any panel and hard to modify the code or unable to access the terminal enough permission then you can perfrom step 1 to 7 at any device, doesnt matter that device is local or online then mention repo URL in your panel just set env variable and you good to go smoothly]

yesbhautik:
BUT BUT BUT, I am getting one another issue. I have white listed one domain because I was getting "Email [email-id] is not allowed to sign up." But i didn't worked for me. I have to white whole email address to get signed up. What if I want to make site open for all with one specified domain or all domains?

ENV:
[...]

PR #1406, opened 2024-02-08
Fix / refactor email whitelist and domain whitelist checks

superhappychris:
This is a fix for issue #1377 

Steps to reproduce: set environment variable values for both EMAIL_DOMAIN_WHITELIST and EMAIL_WHITELIST. Attempt to sign up using an email address that matches only one of the whitelist values

Expected behavior: you are allowed to sign up

Actual behavior: you are blocked from signing up with error message: `Email ${email} is not allowed to sign up.`


fkesheh (approved):
This looks good in my opinion. I tested it on my local, and also run it through AI.

PR #1418, opened 2024-02-10
Added note on entering OpenAI org ID

Itangalo:
I think this issue was what caused problems for me a week ago or so. For experienced users it is probably obvious that the org ID must be set, but it could be worth pointing out anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,9 +133,11 @@ Get the required values by running:
 supabase status
 ```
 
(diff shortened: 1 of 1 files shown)


secondtruth:
OpenAI works for me without entering an Org ID. 🤔

mckaywrigley:
Technically this one is optional but helps to specify for billing purposes

PR #1423, opened 2024-02-10
feat: disable submitting buttons on login page

bmpolonsky:
I encountered a problem with the UI on the login page where the buttons did not show users that their click was registered. I didn't know who to solve it properly, because the page is rendered on the server side. 
To solve this, I decided to disable the buttons after they are clicked. I created a new SubmitButton component that replaces the standard buttons on the login/sign-up form.
After triggering any submit action (sign-in/sign-up/reset), all buttons are disabled.
This at least provides some clear feedback to users by showing that their action is being processed.


fkesheh (commented):
LGTM, very useful

mckaywrigley:
very nice solution

PR #1425, opened 2024-02-11
Allow inactive chats to be renamed and deleted via sidebar buttons

ochen1:
This is useful as loading a new chat currently takes over 20 HTTP requests over the course of 2000 ms for me.
Maybe I will try to figure out what's causing this delay soon.
image

I've made it so that the buttons are displayed on hover:
image


fkesheh (approved):
LGTM

PR #1445, opened 2024-02-15
fix: enable mask for pwa on macos

bmpolonsky:
image

It's disturbing to see squared icon in PWA mode :D
Here's a small tweak to fix it


bmpolonsky:
the result:
image

mckaywrigley:
much better haha

PR #1193, opened 2024-01-16
Local Docker Deployment & Instructions

davidamacey:
## Docker Deployment

In order to serve this application on my system I prefer Docker Instances for ease of deployment.  In an order to setup easily, I created a Makefile, Scripts, and commands to run from the console to increase deployment flexibility.

This includes a docker build file for the chatbot-ui, as well as a docker-compose file for deployment of the UI and ollama.

I did my best to model your setup from the legacy branch.

### Makefile

Using a Makefile, we can easily run commands to deploy, stop, and remove all or parts of the deployment.

### Scripts

New scripts folder includes automatic ollama install and Supabase cli for ease of use.  These may need to be updated in the future for versions and their repo updates.  In the meantime, they will do the job.

### Readme Updated

Updated instructions for using the Makefile with notes necessary for deployment.

### Issue #1088 Bug

In order to get the correct output I added the `streaming: false` parameter per Issue post.  The output is correct, we just don't see the streaming output.  

### Notes

I did my best to streamline the local deployment.  I am open to suggestions and corrections if someone sees a better process or method.

I tested the docker deployment on Ubuntu 22.04 with GPUs.

### Thank you!

Your hardworking on this repo is amazing and very much appreciated!

PR #1476, opened 2024-02-23
Add simple dockerfile

guenhter:
Simple support for running software locally in Docker.

Other then https://github.com/mckaywrigley/chatbot-ui/pull/1193 this PR not supports bringing up the DB and Node-App within on command. This PR just gives the possibility to not startup node locally but within the container. The DB handling is untouched - it must be started before node as usual.


Aj7Ay:
Is this working on instance 
without starting supabase how can we use it?

I tried with your steps and its not working

mckaywrigley:
We're bundling everything soon here

PR #1480, opened 2024-02-23
fix text completion bug

im-calvin:
while the llm is generating text and you type text into the text input, when the llm finishes, it would copy the initial query back into the text input box, removing what you wrote. this fixes that.


vood:
You have an error an API side, this is why you get text back to the input. The original code is correct

PR #1488, opened 2024-02-26
introducing frontend tests in playwright

shray-sharma-phrase:


sharma-shray:
sorry for the missing description. I am pushing some playwright front end tests if you would be interested in having some . we can work on them and make it more exhaustive over time.

PR #1492, opened 2024-02-27
Introducing frontend tests using playwright

sharma-shray:
just to keep things in check some front end tests might come in handy. 
I will be expanding them as we go , creating a pull request for you guys to comment on on the way to expansion of the pack


mckaywrigley:
I'm down to experiment with this!

PR #1508, opened 2024-03-01
Add Google Auth and email verification options to login and env

fkesheh:
Add option for Google Auth
Add option for email verification


mckaywrigley:
Holding off on merging for now because I'm doing a RSC migration and auth changes a bit. Will add and convert as needed once that's done.

PR #1512, opened 2024-03-02
Tools API: Prevent second OpenAI call when no tools used

Gerosullivan:
Opt to return the initial model response directly.

During testing, it was observed that if the OpenAI model provides a high-quality response without utilising any tools, a subsequent call to the model tends to yield a significantly reduced quality in the "final" answer. To enhance efficiency and maintain response quality, this update avoids an unnecessary second invocation of the OpenAI model when no tools are engaged in the process.

### DIFF
--- a/app/api/chat/tools/route.ts
+++ b/app/api/chat/tools/route.ts
@@ -69,6 +69,14 @@ export async function POST(request: Request) {
     messages.push(message)
     const toolCalls = message.tool_calls || []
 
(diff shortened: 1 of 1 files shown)


Gerosullivan:
The if statement in subsequent block (`if (toolCalls.length > 0) {...}` might not be necessary if this is accepted.

mckaywrigley:
Perfect!

PR #1536, opened 2024-03-06
Added QOL feature for localAI users

brtq-01:
## Description

Added a function that is called upon when opening the Model-ID dropdown, which takes the IP address from the IP address input and sends it to the backend to send a GET request to that address and get the Model-IDs for the dropdown

## Security

The IP address is validated and sanitized, by removing the search parameters and all characters that would imply XSS, before being sent to the backend and the data from the GET request is also sanitized before being sent to the frontend

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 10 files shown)


mckaywrigley:
For now we're going to hold off on the additional complexity added here in the main project

PR #1545, opened 2024-03-09
Add Japanese translation.json

burnworks:
Hello,

I've added a translation file for Japanese and would greatly appreciate your review when you have a moment. Thank you for your time and consideration.

Best regards,


mckaywrigley:
We're changing how we do translation soon so maybe revisit Japanese contributions then! :)

burnworks:
Certainly, I'll be sure to revisit Japanese contributions then!

PR #1547, opened 2024-03-09
Introducing Platform Tools

fkesheh:
In this PR I'm introducing a WebScrapper Tool, and also an extensible tool platform to add many other tools (i.e. from langchain) into the Chatbot-UI

For adding or removing platform tools, consult the [platformToolsList.ts](./lib/platformTools/platformToolsList.ts) document. You simply need to add or remove the tools in the list.

To introduce a new tool, utilize the `PlatformTool` interface and incorporate it into the aforementioned list. For a practical illustration, examine the [webscrapperTool.ts](./lib/platformTools/library/webscrapperTool.ts) example.

image


secondtruth:
Great work! Thank you for sharing.

I noticed "Webscraping" is written with 2 P, but it should be just 1.

fkesheh:
Right. I will fix that. There is also a bug around tools when a non open ai model is selected

fkesheh:
Also relative links need to be fixed

vood:
So, I tested this code, and found that the platform tools mess up with server side functions since the code is shared. I.e. with current architecture it is impossible to invoke any server-side only code within the tools, which is a must for scalable system. 

i.e. i want to be able to call getServerProfile and have access to env variables which is not possible right now since getServerProfile throws ` × You're importing a component that needs next/headers. That only works in a Server Component which is not supported in the pages/ directory.`

fkesheh:
You can't acesss server side environment variables from the frontend. That is a security matter. This code has nothing to do with that.

vood:
@fkesheh, let me rephrase the question, let's assume I need to build a DALL-E tool, which needs access to an environment variable to read credentials, how would I use your platform API to achieve this? 

All tools before your change were executed on the server side in `/api/tools` and now with your change, users cannot add server side code to the platform tools. Seems like this breaks existing behavior?

fkesheh:
The tools are still running of the server side. That wasn't changed. The error you postes is related of using getServerProfile in the pages directory. On pages you must use getProfileByUserId as in profiles.ts

vood:
Try adding `getServerProfile` in your `webScraper` tool and see what happens. You can't execute any server side code inside the platform tool since you use `platformToolDefinitions` in the `pages` and `platformToolDefinitions` import all tools, and the tools may contain server side code, which breaks at compile time.

fkesheh:
Got your point. Perhaps you might inject any server-side arguments into the parsedArgs to avoid that, but it feels like a work around. I don't have time now to investigate a better solution

yesbhautik:
can you send that tools values here, so we can create tool from dashboard

vood:
you cannot create it from the dashboard. You need custom code.

yesbhautik:
> you cannot create it from the dashboard. You need custom code.

Because of inactive community of Chatbot UI, we already have created separate repo with UI improvements & implementation of new features. It is no more possible to add beta code snippet's in our repo.

Can try to build tool that I can import from dashboard, or I can just add the context to Supabase, so All user will get that tool by default.

vood:
Same, please follow our development here https://github.com/writingmate/chatlabs

vood:
Again, it is not possible to create a tool that can be "imported from the dashboard". All platform tools require additional coding.

yesbhautik:
> Again, it is not possible to create a tool that can be "imported from the dashboard". All platform tools require additional coding.

Hey, please check my DM on LinkedIn

yesbhautik:
> Same, please follow our development here https://github.com/writingmate/chatlabs

Hey, please check my DM on LinkedIn

yesbhautik:
> Same, please follow our development here https://github.com/writingmate/chatlabs

@vood  Unable to access this repo.

PR #1549, opened 2024-03-10
Refactor Chat and Tools API

fkesheh:
The idea is to create an abstract layer for the provider and use it in other places like in tools or rewriters, re-rankers, etc.

Ideas are welcome.


vood:
This is crazy big update, would be great if you could chunk it

fkesheh:
Actually this is built on top of the platform tools PR, so that your first merge before this one.

I need to move things around as there is logic code in the APIs endpoints and then they can be reused

mckaywrigley:
We're doing a ton of refactors and the merge conflicts will get a little crazy. I'd hold off on larger PRs for the time being. We can revisit after this weekend.

Issue #1441, opened 2024-02-14
Unable to process files > 4.5Mb


fkesheh:
Vercel functions has a 4.5Mb file size limit - Limiting documents that can be processed /api/retrieval/process

https://vercel.com/docs/limits/overview#serverless-function-payload-size-limit

After failing, it doesn't allow user to add any new file.

mpwall:
Having trouble with files > 15MB, trying to upload 100MB file and have set config.toml to match 100MB value but not having luck when uploading from localhost, getting "Failed to upload" error in UI. Any ideas??

PR #1557, opened 2024-03-12
Add file size limit for uploads and fix error handling in file upload

fkesheh:
- Fix the issues #1441


mpwall:
@mckaywrigley this is causing issue on my side. Is there a way we can upload over 10MB files??

Thanks so much, this is a great community!

spammenotinoz:
Partial Fix: allows upload of files dragging from explorer into chat.
Does not address the issue uploading via the "Create Files" menu, this still results in Zero Rows Returned, regardless of file size. eg: 100k.

spammenotinoz:
Narrowed down further, allows uploading of .docx only in the New File\Sidebar menu, all other formats fail.
No issues with drag\drop via explorer.

PR #1565, opened 2024-03-15
Allow the use of Custom Endpoints\Proxies for OpenAI

spammenotinoz:
Minor update that may help other that wish to use either custom end-points or an API proxy such as Cloudflare AI Gateway.
Apologies if this is inefficient or incomplete I can't code, used GPT-4 to guide me.

Why use an AI Gateway?
- Caching (you need to enable this, but not really beneficial, due to the uniqueness of each query)
- Improved Visibility (potentially a privacy issue for some), but useful during debugging
- Ratelimiting

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -31,3 +31,6 @@ AZURE_EMBEDDINGS_NAME=
 # General Configuration (Optional)
 EMAIL_DOMAIN_WHITELIST=
 EMAIL_WHITELIST=
+
+# (Optional) Only define if you wish to use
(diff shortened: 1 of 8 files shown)


kmdavis:
Another benefit is corporate accounts (e.g. my company provides me with an openai api key, but only via a company proxy)

I also have an implementation of this, but on the profile, rather than directly from process.env
https://gist.github.com/kmdavis/709b5cc34d73f14115fbd6cba19f54a4

PR #1589, opened 2024-03-22
Setup instructions for credentials in DB

pacwoodson:
I propose this changes for another way of configuring local/hosted database, without modifying migration files (which induces a difference in code tree)

The idea is just to change the credentials directly within the function in the supabase studio after initial migration


mckaywrigley:
Looks promising. Will investigate.

fkesheh:
I recommend merging this. It's a cleaner approach and less risky.

PR #1590, opened 2024-03-22
In-browser inference with AI-Mask

pacwoodson:
Adding support for [AI-Mask extension](https://github.com/pacoccino/ai-mask) as a local model provider in addition to ollama

AI-Mask is a wrapper on top of libraries such as [web-llm](https://github.com/mlc-ai/web-llm) and [transformers.js](https://github.com/mlc-ai/web-llm) and enable executing models directly in the browser. The extension caches the models and provide them to whatever web app need them, like this one. 

A live demo of this [is available here](https://chatbot.opac.me) ([Install the extension](https://chromewebstore.google.com/detail/lkfaajachdpegnlpikpdajccldcgfdde) before)


mckaywrigley:
Looks interesting but going to hold off for now.

If this project becomes more popular we'll certainly reconsider support!

pacwoodson:
I was hoping to get AI-mask to be known by integrating in some great open source projects that would benefit from it, like this one. 

Maybe you could elaborate on what would refrain you from integrating it ? Would you have specific requirements to match your expectations ?

PR #1625, opened 2024-03-30
Delete > package-lock.json

yesbhautik:
It creates issues in automated server builds…. Also, package-lock.json is self generated file I don’t think it needed in source code.
PR Linked with #1624


AdiRishi:
This doesn't seem wise. Lock files should be included in git commits to ensure reproducible builds. If you don't include it, dependencies of dependencies can be bumped since npm resolution will take the latest version matching constraints. This often leads to bugs and build issues.

PR #1663, opened 2024-04-12
feat: Update GPT models w/ GPT4 Turbo Vision

murillocamargo:
Add new GPT 4 Turbo Vision: https://platform.openai.com/docs/models/continuous-model-upgrades

I'm getting an error trying to use Vision. Even with the option `imageInput` it's throwing the following error:
**400 Invalid type for 'messages[5].content[1].image_url': expected an object, but got a string instead.**

image

### DIFF
--- a/lib/chat-setting-limits.ts
+++ b/lib/chat-setting-limits.ts
@@ -102,6 +102,12 @@ export const CHAT_SETTING_LIMITS: Record<LLMID, ChatSettingLimits> = {
     MAX_CONTEXT_LENGTH: 4096
     // MAX_CONTEXT_LENGTH: 16385 (TODO: Change this back to 16385 when OpenAI bumps the model)
   },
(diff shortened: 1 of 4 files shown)


tim13246879:
I had the same problem. The fix for me was to go to buildFinalMessages() in lib/build-prompts.ts. 

Instead of this:
`
return {
            type: "image_url",
            image_url: formedUrl
          }
`
do this:
`return {
            type: "image_url",
            image_url: {
              "url": formedUrl
            }
          }
`

goququ:
Hi folks, faced with the same issue with new gpt-4o model. Could you help to fix it? Thanks!

```
400 Invalid type for 'messages[5].content[1].image_url': expected an object, but got a string instead.
```
image

tim13246879:
please see above comment.

Same thing is done here #1721

PR #1672, opened 2024-04-16
update gpt 4 turbo

fbec76:
Updated GPT-4 Turbo from the preview version to the production version.


brajeshyadav84:
are we not supporting gpt-4-turbo-preview?  
or we can add new modal on top gpt-4-turbo and keep preview as well.

lib/models/llm/openai-llm-list.ts
tim13246879, line 1: imageInput should be true as the newest model accepts images.

UPDATED time should also be changed for better documentation.

supabase/migrations/20240125192042_upgrade_openai_models.sql
tim13246879, line 1: As mentioned in #1676, this migration file should not be changed. Instead, new migrations should be applied.

tim13246879 (commented):
This PR pretty much replaces the string "gpt-4-turbo-preview" with "gpt-4-turbo" in the codebase. There are other changes that should be made to ensure a smooth transition to the new model though. These are documented in individual comments below, and at #1663.

HowieG:
Hey @fbec76 , like @tim13246879 seems to explain, I had an issue caused by this change. Chatbot UI was not working at all for me. Checked the commit history and saw this change. I put two and two together and changed my selected model from gpt-4-turbo-preview to gpt-4-turbo and it worked. Here's a replay of what happened. After I changed it gpt-4-turbo-preview  was no longer an option so it must've been cached or something, or maybe what @tim13246879 says. Don't have too much time to look into it now but this recording should be helpful.

https://jam.dev/c/534fea88-6411-41ed-b9eb-9b95f47b7fcf

fbec76:
@tim13246879 ,
you are right with all of your points.
We should revert this MR and add the production model as a new model to the preview one.
So we ensure the support of preview chats and functionality.

So, @mckaywrigley please revert this MR!

Edit: The Revert is mentioned in this MR: https://github.com/mckaywrigley/chatbot-ui/pull/1678

PR #1678, opened 2024-04-18
Revert "update gpt 4 turbo"

fbec76:
Reverts mckaywrigley/chatbot-ui#1672

The preview model should not be replaced but the openais models list should be extended with the production gpt-4-turbo


mmahmad:
@mckaywrigley Could we please get this merged to roll back the regressions?

PR #1724, opened 2024-05-17
feat: specify model by query parameter

tomoyuki28jp:
# examples
- http://localhost:3000/ja/xxxxx/chat?model=gpt-4
- http://localhost:3000/ja/xxxxx/chat?model=claude-3-opus-20240229


tomoyuki28jp:
Should I verify if the model is valid?

PR #1792, opened 2024-06-20
Bugfix: key can be caps when shift pressed

FranciscoMoretti:
Compares key pressed for hotkeys **after converting to lowercase**.  Key pressed was received as Caps on my setup

Browser: Chrome
OS: Windows

 Since hotkey requires `shift` to be pressed the key pressed is interpreted as caps. Example, `cmd+shift+o` -> key: `O`


tim13246879:
Great UX improvement to make keyboard shortcuts available.

Issue #1802, opened 2024-07-02
Inline Latex can not show correctly


ukiyo-dev:
![1719892907264](https://github.com/mckaywrigley/chatbot-ui/assets/36793077/e31e55c3-6f4f-43b9-bead-84bcd4fc7a04)

ukiyo-dev:
Raw content is: 
Great question! Let's delve into why \(|z|^n = |z^n|\) for a complex number \(z\) and a positive integer \(n\).

First, recall that for a complex number \(z = a + bi\), where \(a\) and \(b\) are real numbers, the magnitude (or modulus) \(|z|\) is defined as:

\[ |z| = \sqrt{a^2 + b^2} \]

Now, let's consider the complex number \(z\) raised to the power of \(n\):

\[ z^n = (a + bi)^n \]

The magnitude of \(z^n\), denoted as \(|z^n|\), is the modulus of the complex number \(z^n\).

One of the key properties of the modulus of a complex number is that it is multiplicative. This means that for any two complex numbers \(z_1\) and \(z_2\):

\[ |z_1 \cdot z_2| = |z_1| \cdot |z_2| \]

Using this property, we can extend it to \(n\) factors of the same complex number \(z\):

\[ |z^n| = |z \cdot z \cdot \ldots \cdot z| \quad \text{(n times)} \]

Applying the multiplicative property of the modulus repeatedly, we get:

\[ |z^n| = |z| \cdot |z| \cdot \ldots \cdot |z| \quad \text{(n times)} \]

This simplifies to:

\[ |z^n| = (|z|)^n \]

So, we have shown that:

\[ |z^n| = |z|^n \]

This equality holds because the modulus of a product of complex numbers is the product of their moduli, and this property extends naturally to any integer power of a complex number.

PR #1857, opened 2024-09-24
feat: support math formula

### LINKED ISSUE #1223: Cannot convert inline math fomula
I have deployed it to Azure and am using it.
There is an issue where inline formulas are not converted correctly.
I would like to try to fix it myself, but if you know the cause, it would be helpful if you could fix it. 

[Evidence]
ChatGPT
image

Gemini
image

I think other models can occur same problem

usoonees:
fix: https://github.com/mckaywrigley/chatbot-ui/issues/1223 https://github.com/mckaywrigley/chatbot-ui/issues/1802

1. add rehypeKatex plugin to support latex syntax.
2. replace `\(..\)` and `\[...\]` to `$...$` and `$$...$$`. 

Due to remark-math only supporting the use of the `$` symbol for inline math  and `$$` for block math (see discussion at: https://github.com/remarkjs/remark-math/issues/39), we need to convert math notations from `\(..\)` to `$...$` and from `\[...\]` to `$$...$$` .

After:
Conversion result

Before
Appearance before conversion

### DIFF
--- a/app/[locale]/layout.tsx
+++ b/app/[locale]/layout.tsx
@@ -10,6 +10,7 @@ import { Inter } from "next/font/google"
 import { cookies } from "next/headers"
 import { ReactNode } from "react"
 import "./globals.css"
+import "katex/dist/katex.min.css"
 
 const inter = Inter({ subsets: ["latin"] })
 const APP_NAME = "Chatbot UI"
--- a/components/messages/message-markdown.tsx
+++ b/components/messages/message-markdown.tsx
@@ -1,6 +1,7 @@
 import React, { FC } from "react"
 import remarkGfm from "remark-gfm"
 import remarkMath from "remark-math"
+import rehypeKatex from "rehype-katex";
 import { MessageCodeBlock } from "./message-codeblock"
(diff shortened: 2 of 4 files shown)


xl0:
@usoonees I tried this approach for my project, don't think this is the way to go, there are too many corner-cases.
I think we should write a replacement for `remark-math` / `mdast-util-math` to handle them properly.

xl0:
For example:
- LLMs that use the bracket syntax will not escape $ in text, so `$100 to $200` will be seen as a formula.
- LLMs will produce `\[ ... \]` without new lines, and use syntax that is only allowed in block math inside. This is parsed as inline math, and gives a rendering error.
- Math inside
  -  Deeply
      - Nested
          - Lists

requires special handling, as
```
      $$
      \formula
$$
```
is not handled correctly.

Another concern, this project is kinda dead, if you consider the number of open PRs and lack of action. I used to host chatbot-ui for myself and friends, but am now working on [Congusto Chat](https://chat.congusto.ai), which will soon reach feature parity, and has an ambitious roadmap, including code interpreter and artefacts.

BorislavStefanov31 (approved):

PR #1929, opened 2025-04-19
Fix/skip tests for deploy

Saint-Visions:
Temporarily skipping build & test steps to allow full deployment on Azure. Will re-enable after initial deployment is confirmed successful.


Saint-Visions (commented):
“Push skip-tests now Sal”

